PR #9, opened 2022-12-09
Use deployment target 11 as stated in the docs.

yunikkk:


serraojoao:
Hi, is this update being release soon? 

I was just testing on and iOS device and i needed the minimum target version as 15.0.

PR #39, opened 2023-01-13
Expose attribution/logo/compass/scalbar controls [MAPSFLT-86].

yunikkk:


android/src/main/kotlin/com/mapbox/maps/mapbox_maps/AttributionController.kt
```diff
@@ -0,0 +1,15 @@
+package com.mapbox.maps.mapbox_maps
+
+import com.mapbox.maps.MapView
+import com.mapbox.maps.mapbox_maps.mapping.applyFromFLT
+import com.mapbox.maps.mapbox_maps.mapping.toFLT
+import com.mapbox.maps.pigeons.FLTSettings
+import com.mapbox.maps.plugin.attribution.attribution
+
+class AttributionController(private val mapView: MapView) :
```
alexander-kulikovskii, line 9: nit:
maybe not necessary to pass whole mapView? only `attribution` and `context`

android/src/main/kotlin/com/mapbox/maps/mapbox_maps/MapboxMapController.kt
```diff
@@ -79,6 +85,9 @@ class MapboxMapController(
     FLTSettings.LocationComponentSettingsInterface.setup(messenger, null)
     FLTSettings.LogoSettingsInterface.setup(messenger, null)
     FLTSettings.GesturesSettingsInterface.setup(messenger, null)
+    FLTSettings.CompassSettingsInterface.setup(messenger, null)
```
alexander-kulikovskii, line 88: nit:
maybe add to some list and call `setup` in loop?

example/lib/ornaments.dart
```diff
@@ -0,0 +1,285 @@
...
+    );
+  }
+
+  Widget _moveOrnaments() {
+    return TextButton(
+      child: Text('move compass'),
+      onPressed: () {
+        setState(() {
+          switch (compassPosition) {
+            case OrnamentPosition.BOTTOM_LEFT:
```
alexander-kulikovskii, line 141: nit:
make some map of `current position` -> `new position` and set new position in one line

alexander-kulikovskii (approved):

android/src/main/kotlin/com/mapbox/maps/mapbox_maps/AttributionController.kt
yunikkk: I'd left as is - it's implemented similarly within all controllers

android/src/main/kotlin/com/mapbox/maps/mapbox_maps/MapboxMapController.kt
yunikkk: Not possible since `setup` is not an interface method but static one implemented for every generated controller by Pigeon

example/lib/ornaments.dart
yunikkk: IMO not worth it, enum isn't that long and switch is neither, lets keep it

yunikkk:
Firebase struggle to finish Android tests for some reason even after numerous restarts, though locally they work fine. Merging this one and will deal with Firebase debug later.

PR #65, opened 2023-02-03
Export package turf.

RennanAgroSmart:
It is needed export the turf package, for use the **Postion** class and others. Otherwise, this package can't be used with the turfs types.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.4.1
+
+### Common
+* Expose `package:turf/helpers.dart`.
+
 ## 0.4.0 
 
 ### Common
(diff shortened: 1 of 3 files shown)


yunikkk (approved):

LjubiTech-Maxko:
@yunikkk This export should be moved before the part directives.

yunikkk:
> @yunikkk This export should be moved before the part directives.

Won't it be recognized otherwise? I'll check thanks.

LjubiTech-Maxko:
I haven't tested because I'm using turf package as a dependency, but I can't build the app with this as the following error is present:

```
Export directives must precede part directives.  
Try moving the export directives before the part directives.
```

Issue #68, opened 2023-02-06
How to keep MapWidget() from recreating on every navigation back to the Widget ?


Shiba-Kar:
The ```MapWidget()``` recreate on every navigation back to the Screen where there is ```MapWidget()``` as a result ```onMapCreated``` is called  where the ```style.setStyleURI```  is called and there is jank . How can I prevent this jank ?

Issue #57, opened 2023-01-24
onMapCreated is called when app is switching to background


kamami:
When putting the application in the background and switch to another app, onMapCreated is called. Indeed it is called twice...

```
  _onMapCreated(MapboxMap mapboxMap) async {
    print("onMapCreated");
}
```

**EDIT:**

```
import 'package:flutter/material.dart';
import 'package:mapbox_maps_flutter/mapbox_maps_flutter.dart';

void main() async {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  @override
  Widget build(BuildContext context) {
    return const MaterialApp(
      home: MapView(),
    );
  }
}

class MapView extends StatefulWidget {
  const MapView({super.key});

  @override
  State<MapView> createState() => _MapViewState();
}

class _MapViewState extends State<MapView> {
  
  _onMapCreated(MapboxMap mapboxMap) async {
    print("onMapCreated");
  }

  @override
  Widget build(BuildContext context) {
    return MapWidget(
      cameraOptions: CameraOptions(
          padding: MbxEdgeInsets(
              bottom: (MediaQuery.of(context).size.height -
                      kBottomNavigationBarHeight -
                      MediaQuery.of(context).padding.top) *
                  0.15,
              left: 0,
              right: 0,
              top: 0),
          zoom: 4,
          center: {
            "coordinates": [9.142084916929944, 49.97744579860142]
          }),
      resourceOptions: ResourceOptions(
          accessToken:
              "XXX"),
      onMapCreated: _onMapCreated,
    );
  }
}
```

kamami:
@LjubiTech-Maxko I edited the initial issue. 
Seems to be related to the CameraOptions imo.

LjubiTech-Maxko:
Please edit the sample code and delete your accessToken.

kamami:
> Please edit the sample code and delete your accessToken.

Yes, just saw it. Thanks

LjubiTech-Maxko:
This is probably related to the #26. Every time build is called (regardless of the `MediaQuery` usage in the `CameraOptions`) on the widget which contains the `MapWidget`, `onMapCreated` is called which can be seen in the following source code from repo.

https://github.com/mapbox/mapbox-maps-flutter/blob/308d25dac22f2402a79292cd6fa5a37f2973643b/lib/src/map_widget.dart#L177-L195

[...]

Issue #26, opened 2023-01-04
Map is rerendering every frame when keyboard appears/disappears


kamami:
The `_onMapCreated` function is triggered every single frame, when the soft keyboard is appearing or disappearing.

```
_onMapCreated(MapboxMap mapboxMap) async {
    print("ONMAPCREATED");
//Prints to console every single frame
}
```

Console:

flutter: ONMAPCREATED
flutter: ONMAPCREATED
flutter: ONMAPCREATED
flutter: ONMAPCREATED
flutter: ONMAPCREATED
flutter: ONMAPCREATED
flutter: ONMAPCREATED
flutter: ONMAPCREATED
flutter: ONMAPCREATED
flutter: ONMAPCREATED
flutter: ONMAPCREATED
flutter: ONMAPCREATED
flutter: ONMAPCREATED
flutter: ONMAPCREATED
flutter: ONMAPCREATED
flutter: ONMAPCREATED

dawid-niedzwiecki:
Sorry, made a typo in my issue :/

PR #71, opened 2023-02-07
Fix mapCreated callback called on every build

LjubiTech-Maxko:
# Description

Considering flutters posibility for calling build method on every frame as stated in the docs,

> This method can potentially be called in every frame and should not have any side effects beyond building a widget.

we can say that `MapWidget`s build method has side effects of calling `onMapCreated` callback which should be called only the first time map is created, which is already done from `onPlatformViewCreated` callback.

Fixes: #26 
Fixes: #57 
Fixes: #68

## Type of change

- [x] Bug fix (non-breaking change)

### DIFF
--- a/lib/src/map_widget.dart
+++ b/lib/src/map_widget.dart
@@ -186,10 +186,6 @@ class _MapWidgetState extends State<MapWidget> {
       'mapboxPluginVersion': '0.4.0'
     };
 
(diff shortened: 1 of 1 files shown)


yunikkk (approved):
@LjubiTech-Maxko LGTM, thanks for the contribution!

PR #76, opened 2023-02-08
Fix getLayer on layers with expression properties. 

### LINKED ISSUE #14: SymbolLayer bad decoding of expression properties
Using `mapboxMap.style.getLayer()` will throw exception if layer properties are using expressions. In the sample code bellow, which is using default mapbox light style, text-size property is using the following expression, which is `List<dynamic>` when decoded, but `double` is expected.

```
"text-size": [
    "step",
    ["zoom"],
    ["step", ["get", "sizerank"], 18, 5, 12],
    17,
    ["step", ["get", "sizerank"], 18, 13, 12]
],
```

Reproduction code:

```
void main() {
  runApp(const MyApp());
}

class MyApp extends StatefulWidget {
  const MyApp({super.key});

  @override
  State<MyApp> createState() => _MyAppState();
}

class _MyAppState extends State<MyApp> {
  late final MapboxMap _map;

  void _mapCreated(MapboxMap map) {
    _map = map;
  }
  
  void _onStyleLoaded(StyleLoadedEventData event) async {
    await _map.style.getLayer('natural-point-label');
  }

  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      title: 'Flutter Get Symbol Layer Exception',
      theme: ThemeData(
        primarySwatch: Colors.blue,
      ),
      home: Scaffold(
        body: MapWidget(
          resourceOptions: ResourceOptions(accessToken: <add access token>),
          styleUri: MapboxStyles.LIGHT,
          onMapCreated: _mapCreated,
          onStyleLoadedListener: _onStyleLoaded,
        ),
      ),
    );
  }
}
```

When running this code the following exception is thrown:

```
E/flutter (22989): [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: NoSuchMethodError: Class 'List<dynamic>' has no instance method 'toDouble'.
E/flutter (22989): Receiver: Instance(length:5) of '_GrowableList'
E/flutter (22989): Tried calling: toDouble()
E/flutter (22989): #0      Object.noSuchMethod (dart:core-patch/object_patch.dart:38:5)
E/flutter (22989): #1      SymbolLayer.decode (package:mapbox_maps_flutter/src/style/layer/symbol_layer.dart:580:45)
E/flutter (22989): #2      StyleLayer.getLayer (package:mapbox_maps_flutter/src/style/style.dart:217:29)
E/flutter (22989): <asynchronous suspension>
E/flutter (22989): #3      _MyAppState._onStyleLoaded (package:mapbox_get_layer_exception/main.dart:23:5)
E/flutter (22989): <asynchronous suspension>
E/flutter (22989): 
```

I think this is a major issue because we can add our own layers which are using expressions but can't retrieve the same layer later.

yunikkk:
Properties that are set to expressions will return nulls. 
To retrieve such properties lower level style apis can be used at the moment, e.g. : `map.style.getStyleLayerProperty(LAYER, PROPERTY);`

Fixes #14 / MAPSFLT-89, cc @LjubiTech-Maxko

### DIFF
--- a/lib/src/style/layer/background_layer.dart
+++ b/lib/src/style/layer/background_layer.dart
@@ -80,8 +80,8 @@ class BackgroundLayer extends Layer {
               .toLowerCase()
               .contains(map["layout"]["visibility"])),
       backgroundColor: (map["paint"]["background-color"] as List?)?.toRGBAInt(),
-      backgroundOpacity: map["paint"]["background-opacity"] is double
-          ? map["paint"]["background-opacity"] as double
+      backgroundOpacity: map["paint"]["background-opacity"] is double?
+          ? map["paint"]["background-opacity"] as double?
           : null,
       backgroundPattern: map["paint"]["background-pattern"],
     );
--- a/lib/src/style/layer/background_layer.dart
+++ b/lib/src/style/layer/background_layer.dart
@@ -80,7 +80,9 @@ class BackgroundLayer extends Layer {
               .toLowerCase()
               .contains(map["layout"]["visibility"])),
       backgroundColor: (map["paint"]["background-color"] as List?)?.toRGBAInt(),
-      backgroundOpacity: map["paint"]["background-opacity"]?.toDouble(),
+      backgroundOpacity: map["paint"]["background-opacity"] is double
+          ? map["paint"]["background-opacity"] as double
+          : null,
       backgroundPattern: map["paint"]["background-pattern"],
     );
   }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * Expose `package:turf/helpers.dart`.
 * Bump platform Maps SDK dependencies to 10.10.0.
 * Fix issue with multiple maps overriding platform channels of the previous instances.
+* Fix exception accessing `style.getLayer` when layer property is an Expression.
 
 ## 0.4.0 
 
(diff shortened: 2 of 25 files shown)


lib/src/style/layer/background_layer.dart
```diff
@@ -80,7 +80,9 @@ class BackgroundLayer extends Layer {
               .toLowerCase()
               .contains(map["layout"]["visibility"])),
       backgroundColor: (map["paint"]["background-color"] as List?)?.toRGBAInt(),
-      backgroundOpacity: map["paint"]["background-opacity"]?.toDouble(),
+      backgroundOpacity: map["paint"]["background-opacity"] is double
```
alexander-kulikovskii, line 83: just a question. is it possible to write some extensions like in kotlin?

```
getDoubleOrNull(map["paint"]["background-opacity"]){
  if this is double return this else null
}
```

alexander-kulikovskii (approved):

lib/src/style/layer/background_layer.dart
yunikkk: Yeah we can add an extension but I wouldn't optimize it since it's just a generated code

yunikkk:
> I noticed that in decode method of layers proposed changes are all getting properties from "paint" property of map but some were previously in "layout" property.
> 
> For example in symbol layer:
> 
> ```dart
> -      textSize: map["layout"]["text-size"]?.toDouble(),
> +      textSize: map["paint"]["text-size"] is double
> +         ? map["paint"]["text-size"] as double
> +          : null,
> ```


Good catch, wrong codegen update. Fixing.

PR #104, opened 2023-02-28
Fix GestureListener not being disposed

LjubiTech-Maxko:
## Description

There is a memory leak if we add `onTapListener`, `onLongTapListener` or `onScrollListener` callbacks. `GestureListener` is set up but never disposed.

Also, considering the documentation for dispose method which states:
> Implementations of this method should end with a call to the inherited method, as in super.dispose().

I moved those calls to the end of the method.

Edit: Reverted MapWidget super.dispose() call because, I assume, widget is already disposed and unmounted while _controller.future is being awaiting for, which caused the following assertion `_MapWidgetState.dispose failed to call super.dispose.`

## Type of change
- Bug fix (non-breaking change)

### DIFF
--- a/lib/src/map_widget.dart
+++ b/lib/src/map_widget.dart
@@ -197,11 +197,11 @@ class _MapWidgetState extends State<MapWidget> {
 
   @override
   void dispose() async {
-    super.dispose();
     if (_controller.isCompleted) {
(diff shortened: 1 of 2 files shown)


yunikkk:
@LjubiTech-Maxko thanks for contribution!

PR #172, opened 2023-05-24
Fix iOS bearing icon

LesGrob:
Fix iOS bearing icon configuration


yunikkk:
Will land in https://github.com/mapbox/mapbox-maps-flutter/pull/182.

PR #194, opened 2023-06-17
fix: Point to FLT ScreenCoordinate conversion

LorenzSchueler:
This PR fixes #81 and #190 (Android only).

Currently the `onTap`, `onLongTap` and `onScroll` callbacks get a `ScreenCoordinate`. However the `x` and `y` values do not correspond to the pixels on the screen but are actually the latitude and longitude. This is cause by a incorrect conversion in the GestureController, which takes the point (latitude and longitude) values to build the `ScreenCoordinate`. This PR fixes that by converting the point to a `ScreenCoordinate` using `pixelForCoordinate`.

It seems like the ios code suffers from the same problem. However I have no knowledge of Swift and also no means to test a potential fix for ios. Therefore I'm leaving that for someone else to fix.

Since we have the point (latitude and longitude) already, I'm going to create a follow up PR (#195) that additionally to the `ScreenCoordinate` also provides the point (as `Map<String?, Object?>` like in other places) to the callback.

PR #195, opened 2023-06-17
feat: provide point in callbacks

LorenzSchueler:
follow-up PR for #194 
see #194 for more details


evil159:
@LorenzSchueler Thank you for your contribution, we have added both touch position and projected map coordinate for the touch position to the gesture callbacks in https://github.com/mapbox/mapbox-maps-flutter/pull/473. These changes has diverged from the direction of this PR and we decided to start anew in a separate PR with them. Thank you for raising this and providing a suggested solution!

PR #234, opened 2023-09-15
Upgrade Pigeon to v11

evil159:
Mainly this PR updates Pigeon to version `11.0.1` and sequentially updates min dart sdk version to 2.19.0.

Additionally, a few changes were made to address various issues:
* sets `MapboxStyles.MAPBOX_STREETS ` as a default value for `styleUri`(map_widget.dart)
* speeds up `swiftlint` install by disabling Homebrew update(config.yml)
* adds a workaround for failing tests on iOS because of overlaying locations permission dialog(all_test.dart)
* adds a workaround for failing `getDragCameraOptions` test by scaling inputs on Android(camera_test.dart)


ios/Classes/LocationController.swift
```diff
@@ -142,6 +142,19 @@ extension LocationOptions {
             }
         }
 
-        return FLT_SETTINGSLocationComponentSettings.make(withEnabled: enabled, pulsingEnabled: nil, pulsingColor: nil, pulsingMaxRadius: nil, showAccuracyRing: showAccuracyRing, accuracyRingColor: accuracyRingColor, accuracyRingBorderColor: accuracyRingBorderColor, layerAbove: nil, layerBelow: nil, puckBearingEnabled: puckBearingEnabled, puckBearingSource: puckBearingSource, locationPuck: locationPuck)
+        return FLT_SETTINGSLocationComponentSettings.make(
+            withEnabled: enabled,
+            pulsingEnabled: nil,
```
yunikkk, line 147: Dont remember why wasn't pulsing mapped previously, guess it is not available on iOS at all?

example/integration_test/map_interface_test.dart
```diff
@@ -200,11 +200,7 @@ void main() {
 
     var options = await mapboxMap.getResourceOptions();
     expect(options.accessToken, isNotNull);
-    if (Platform.isAndroid) {
-      expect(options.baseURL, 'https://api.mapbox.com');
-    } else {
-      expect(options.baseURL, 'file:///https:/api.mapbox.com');
```
yunikkk, line 206: what was the reason for the file prefix?

evil159: This was caused by a bug in the iOS SDK https://github.com/mapbox/mapbox-maps-ios/pull/1749, which was fixed in 10.11.0.

ios/Classes/LocationController.swift
evil159: Location indicator pulsing is available on iOS - 
https://github.com/mapbox/mapbox-maps-ios/blob/8b25fceb8570b0f9ca829685769396f4e02cd7e0/Sources/MapboxMaps/Location/Puck/PuckType.swift#L65

yunikkk: Ah ok, just missed in the past then, good to know.

Issue #134, opened 2023-04-11
Make the iconColor property for the attribution icon work


ToniTornado:
I would like to change the icon color for the attribution icon from default blue to match my app theming.

Setting the iconColor like this does not seem to work as of now:

```
controller.attribution.updateSettings(
  AttributionSettings(
    iconColor: Colors.white.value,
  ),
)
```

I found this comment in the tests:

```
// FIXME colors are decoded incorrectly for some reason
// expect(updatedSettings.iconColor, Colors.blue.value);
```

arctouch-BrunoFerrari:
Hi, I'm facing the same problem, someone could please prioritize this fix?

I can confirm it works on android but not on iOS on v0.4.4

arctouch-matheusromao:
@yunikkk, any chance you could take a look at this one, please? Thank you!

yunikkk:
Hey everyone, this should be fixed with https://app.circleci.com/pipelines/github/mapbox/mapbox-maps-flutter/146

PR #252, opened 2023-10-03
Fix attribution button color

evil159:
This PR fixes attribution button color not being applied on iOS. As well as, fixes the issue of attribution color not decoding correctly on Android.


yunikkk (dismissed):

yunikkk:
@evil159 could you reformat the project to fix the Android build? There is a make target for that, `make format` should do the job. 

This one also fixes #134

yunikkk (approved):

PR #254, opened 2023-10-03
Batch ornaments settings on iOS

evil159:
Instead of applying ornament options one by one batch them up and apply all at once(per ornament) to avoid unnecessary updates triggered by its setter.


yunikkk:
@evil159 fyi conflicts here

ios/Classes/CompassController.swift
```diff
@@ -4,36 +4,33 @@ import UIKit
...
         if let visible = settings.enabled {
             if !visible.boolValue {
-                mapView.ornaments.options.compass.visibility = OrnamentVisibility.hidden
-            } else {
-                if settings.fadeWhenFacingNorth?.boolValue ?? true {
-                    mapView.ornaments.options.compass.visibility = OrnamentVisibility.adaptive
-                } else {
-                    mapView.ornaments.options.compass.visibility = OrnamentVisibility.visible
-                }
+                compass.visibility = visible.boolValue ? .adaptive : .hidden
```
yunikkk, line 29: Hmm is it correct if's sequence?

evil159: Ah, true, I missed the `fadeWhenFacingNorthPart` 🤦

evil159: Should be better now

yunikkk (approved):
LGTM!

yunikkk:
@evil159 auto rebase does not pass, 
```
Update branch attempt failed
Rebase conflict between head and base.
```
guess needs manual

PR #256, opened 2023-10-04
Update CI config

evil159:
This PR introduces following changes to the CircleCI config:
* update macOS runner to M1: `macos.m1.medium.gen1`
* update Xcode version to `15.0.0`
* use `circleci/macos` - mainly because it makes it easy to install Rosetta which is required for Flutter on M1
* use `gcp-cli` orb for `gcloud`(also update to the latest version: `449.0.0`)
* bump Flutter SDK version to `3.13.6`
* kick off firebase tests and wait the results on a small docker image(should be a bit cheaper I guess)
* remove duplicate code signing and build steps for iOS

This effectively halves iOS build time(~-50%), so now it is on par with Android(~5m), which results in a shorted turnaround time for the whole workflow(~10m now vs ~16m before).
Plus, saves a bit of time installing `glcould`(not sure how but the orb installs it faster than when it was done "manually").

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,8 +67,7 @@ jobs:
       workspace-path:
         type: string
         default: example/build/app/outputs/apk
-    docker:
-      - image: cimg/python:3.7    
+    executor: firebase-test-runner
     steps:
       - attach_workspace:
           at: << parameters.workspace-path >>
@@ -135,8 +134,7 @@ jobs:
       workspace-path:
         type: string
         default: example/build/output
-    docker: 
-      - image: cimg/python:3.7
+    executor: firebase-test-runner
     steps:
       - attach_workspace:
           at: << parameters.workspace-path >>
@@ -168,3 +166,7 @@ executors:
     resource_class: macos.m1.medium.gen1
     macos:
       xcode: 15.0.0
+  firebase-test-runner:
+    resource_class: small
+    docker:
+      - image: cimg/python:3.7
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -142,15 +111,11 @@ jobs:
       - run:
           name: "Build iOS integration tests"
           command: |
-            bundle exec fastlane update_code_signing
-
             pushd example
 
             # required to init configs that pass environment variables to fastlane later
             flutter build ios --config-only integration_test/all_test.dart \
```
yunikkk, line 117: @evil159 hmm and is it enough, build still succeeds with the single command? AFAIR previously it was needed to first build with `--config-only` flag and only after that build it actually...

evil159: @yunikkk this is exactly what is happening now - first we update the config with `flutter build ios --config-only ...` and then build it with `bundle exec fastlane build_examples_tests`

yunikkk: The instructions are here https://github.com/flutter/flutter/tree/main/packages/integration_test#ios-device-testing, it mentions config-only but is actually quite confusing whether it is mandatory step...

yunikkk (approved):
Few questions but LGTM since the build passes

.circleci/config.yml
yunikkk: > @yunikkk this is exactly what is happening now - first we update the config with `flutter build ios --config-only ...` and then build it with `bundle exec fastlane build_examples_tests`

I see. Not really sure why that second build was here, probably from the times before fastlane was added

PR #307, opened 2023-11-24
feature: Physical position on a map is available by pixel in the callbacks of onTap/onLongTap/onScroll

toriumi0118:
It is inspired, although I reused it, by https://github.com/mapbox/mapbox-maps-flutter/pull/194 .

I was struggling that it was impossible to obtain a position which point of device is tapped physically. The positions, however, are needed to use the `queryRenderedFeatures` method.

Therefore, I added the implementation of iOS that is counterpart of android, which is already there.

Indeed, it already enabled to get geo location by lat/lng though, it looks wired that the values derive from coordinate `x/y` that should have been `lat/lng` or something like it to avoid misreading. Notwithstanding, I do not want to change these interface, so I endorse current style to add `position` parameter to the callbacks.

In addition, https://github.com/mapbox/mapbox-maps-flutter/pull/194 is conflicted to main branch, hence I also resolved it.

### DIFF
--- a/android/src/main/java/com/mapbox/maps/pigeons/FLTGestureListeners.java
+++ b/android/src/main/java/com/mapbox/maps/pigeons/FLTGestureListeners.java
@@ -180,28 +180,28 @@ public class FLTGestureListeners {
     static @NonNull MessageCodec<Object> getCodec() {
       return GestureListenerCodec.INSTANCE;
     }
-    public void onTap(@NonNull ScreenCoordinate coordinateArg, @NonNull Reply<Void> callback) {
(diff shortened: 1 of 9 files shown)


evil159:
@toriumi0118 Thank you for your contribution, we have added both touch position and projected map coordinate for the touch position to the gesture callbacks in https://github.com/mapbox/mapbox-maps-flutter/pull/473. These changes has diverged from the direction of this PR and we decided to start anew in a separate PR with them. Thank you for raising this and providing a suggested solution!

toriumi0118:
@evil159 
Thank you for your update.
It sounds nice, and I suppose it is better way to solve the problem.

I have thought that almost all information (data) derived from native SDK should be returned to the flutter code.
Therefore, using a context is reasonable idea.

By the way, I checked the issues and found the alternative way to solve it. I have used pixelForCoordinate instead lately.

PR #317, opened 2023-12-08
[v11] resolve build errors iOS

maios:
1fdd691652e9020c7bbaac946eacd6c1231072dd Convert `StyleColor` to integer representation. In v11, `StyleColor` supports RGB(A), HSL(A), hex, and a raw string as shown [here](https://docs.mapbox.com/style-spec/reference/types/#color) for a moment, we support HSL(A) and RGB(A) only.
49605d6f306f89569951871ec16653e47cf2943c Add conformers for `_MapboxOptions` and `_MapboxMapsOptions` that is responsible for managing access token and map's configuration (previously known as ResourceOptionsManager).

Upstream https://github.com/mapbox/mapbox-maps-flutter-internal/pull/126

### DIFF
--- a/ios/Classes/MapEvents+JSON.swift
+++ b/ios/Classes/MapEvents+JSON.swift
@@ -9,8 +9,8 @@ protocol MapEventEncodable {
 extension EventTimeInterval {
     var toJSON: [String: Any] {
         [
-            "begin": begin.milisecondsSince1970,
-            "end": end.milisecondsSince1970
+            "begin": begin.microsecondsSince1970,
+            "end": end.microsecondsSince1970
         ]
     }
 }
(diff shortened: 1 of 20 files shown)


ios/Classes/MapEvents+JSON.swift
```diff
@@ -9,8 +9,8 @@ protocol MapEventEncodable {
 extension EventTimeInterval {
     var toJSON: [String: Any] {
         [
-            "begin": begin,
-            "end": end
+            "begin": begin.timeIntervalSince1970,
```
evil159, line 12: I think on the flutter side this is represented in microseconds, but his is in seconds.

maios: ah I just know it, will fix it soon

ios/Classes/MapEvents+JSON.swift
```diff
@@ -9,8 +9,8 @@ protocol MapEventEncodable {
 extension EventTimeInterval {
     var toJSON: [String: Any] {
         [
-            "begin": begin,
-            "end": end
+            "begin": begin.milisecondsSince1970,
```
evil159, line 12: Flutter docs state microseconds for the timestamp, maybe we could try using DateTime instead? Seems that it's already confusing for us(milli/micro), probably will be more some to the users.

evil159 (approved):

PR #322, opened 2023-12-13
Registry token lookup

evil159:
Fix registry token lookup failing with an unrelated error in certain circumstances.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -15,7 +15,7 @@ buildscript {
 }
 
 rootProject.allprojects {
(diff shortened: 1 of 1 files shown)


maios (approved):
btw could you explain how is it different for the change? Why does `project.properties` subscript not work and `project.findProperty` work?

evil159:
> btw could you explain how is it different for the change? Why does `project.properties` subscript not work and `project.findProperty` work?

Not sure exactly, but I suspect the error could be caused by the fact `project.properties('...')` throws an exception if a property is not found, `project.findProperty('...')` doesn't throw an exception in this case, simply returning `null`.

PR #325, opened 2023-12-13
[v11] Fix integration tests

maios:
- Fix failing integrations
- Support Lights API for iOS
Upstream PR https://github.com/mapbox/mapbox-maps-flutter-internal/pull/129

**TODO**

- [ ] [Write tests for MapEvents](https://mapbox.atlassian.net/browse/MAPSFLT-139)
- [ ] Investigate and update GeoJSONSource integration tests


example/integration_test/camera_test.dart
```diff
@@ -155,14 +147,8 @@ void main() {
...
             min: ScreenCoordinate(x: 0, y: 0),
             max: ScreenCoordinate(x: 100, y: 100)));
-    expect(camera.zoom, 10);
     expect(camera.bearing, 20);
     expect(camera.pitch, 30);
-    expect(camera.padding!.top, 1);
-    expect(camera.padding!.left, 2);
-    expect(camera.padding!.bottom, 3);
-    expect(camera.padding!.right, 4);
-    expect(camera.zoom!.round(), 10);
```
evil159, line 165: Why these are not being checked anymore?

maios: these camera API won't return a padding, at the moment it seems like only `cameraForCoordinates:camera:coordinatesPadding:maxZoom:offset` will return a padding value.

evil159: I think in this case returned camera options `padding` should be the same as the `padding` provided by supplied camera options, thus we can re-instantiate padding checks.

evil159: Additionally, zoom check should also be restored, right?

evil159 (approved):

PR #335, opened 2023-12-22
Fix integration tests

evil159:


ios/Classes/MapboxMapFactory.swift
```diff
@@ -76,41 +76,41 @@ class MapboxMapFactory: NSObject, FlutterPlatformViewFactory {
...
-            let bearing: CLLocationDirection? =  cameraOptionsMap[4] as? CLLocationDirection
-            let pitch: CGFloat? =  cameraOptionsMap[5] as? CGFloat
-
-            if let centerMap = cameraOptionsMap[0] as? [String: Any] {
-                center = convertDictionaryToCLLocationCoordinate2D(dict: centerMap)
-            }
+    private func createCameraOptions(args: [String: Any]) -> CameraOptions? {
+        guard let cameraOptionsMap = args["cameraOptions"] as? [Any] else {
+            return nil
+        }
```
evil159, line 82: Git shows a bunch of changes, but the only thing I did was to make the return type optional and add early return in case camera options were not provided.
(This is needed because even empty camera options `CamerOptions()` seem to override style default camera).

maios (approved):

PR #358, opened 2024-01-12
Make all methods of MapboxMapsOptions static

evil159:
This PR converts all getters/setters in `MapboxMapsOptions` to be static(it was overlooked). Plus, makes constructors of `MapboxMapsOptions` and `MapboxOptions` private, as these classes are not supposed to be initialized.

### DIFF
--- a/lib/src/mapbox_maps_options.dart
+++ b/lib/src/mapbox_maps_options.dart
@@ -1,7 +1,7 @@
 part of mapbox_maps_flutter;
 
 /// Class for Mapbox SDK settings management
-class MapboxOptions {
+final class MapboxOptions {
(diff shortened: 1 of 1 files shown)


lib/src/mapbox_maps_options.dart
```diff
@@ -4,6 +4,8 @@ part of mapbox_maps_flutter;
 class MapboxOptions {
```
maios, line 4: can these 2 classes be `final`?

maios (approved):

lib/src/mapbox_maps_options.dart
evil159: Well spotted, made it so 👍

maios (approved):

Issue #377, opened 2024-01-18
[BUG] View blank on Pixel 8 device, under Android 14 system.


figureai:
Like the image, When I run the application on  Pixel 8 device, under Android 14 system. The view was blank. But it runs normally on other Android models and systems.
![IMG_4321](https://github.com/mapbox/mapbox-maps-flutter/assets/45620405/9b0389f0-9daf-4a1d-8a85-09f84ab00933)
This is my code. I have hid the accessToken.
`body: MapWidget(
        resourceOptions: ResourceOptions(
          accessToken:
              "*************************************",
        ),
      ),`
and my pubspec.yaml file.
`mapbox_maps_flutter: ^0.5.1`
I am very confused about this issue and deeply appreciate anyone who can provide a solution. Thank you very much.

baleboy:
Could you check if this is reproducible on 1.0.0-beta.2 ?

wnowak-vialytics:
@baleboy I also got it on 1.0.0-beta.2

I think it's related to reading mapOptions at 
https://github.com/mapbox/mapbox-maps-flutter/blob/main/android/src/main/kotlin/com/mapbox/maps/mapbox_maps/MapboxMapFactory.kt#L40
The code skips one position, moving from `mapOptions[4]` to `mapOptions[6]`:
```
      mapOptions[4]?.let {
        mapOptionsBuilder.crossSourceCollisions(it as Boolean)
      }
      mapOptions[6]?.let {
        (it as ArrayList<Double>).let { size ->
```
I got an error in the console regarding an incorrect casting (from `double` to `ArrayList`). Fixing indices fixed the map, however, no markers are visible (might be unrelated to the issue, I need to investigate more, as I just migrated to the new version).

figureai:
> 您能检查一下这是否可以在 1.0.0-beta.2 上重现吗？

Yes, In version 1.0.0-beta.2, the same issue of a blank view. I find it only occurs on Pixel 8 device.
![IMG_4321](https://github.com/mapbox/mapbox-maps-flutter/assets/45620405/0fed93ae-dc27-4a09-89c1-4f776b88cb02)

baleboy:
thanks for the extra details. I think the description of the issue is quite clear, but FYI  the screenshot is not visible.

figureai:
> thanks for the extra details. I think the description of the issue is quite clear, but FYI the screenshot is not visible.

[...]

PR #379, opened 2024-01-23
[Android] Fix MapOptions incorrect index access at map creation

LjubiTech-Maxko:
### Fixes map not beeing created (blank map) issues.

As pointed out in the https://github.com/mapbox/mapbox-maps-flutter/issues/377#issuecomment-1904311347, incorrect index access to mapOptions in `MapboxMapFactory` create method could be causing blank map issues.

The following exception would be thrown:

`Unhandled Exception: PlatformException(error, java.lang.Double cannot be cast to java.util.ArrayList, null, java.lang.ClassCastException: java.lang.Double cannot be cast to java.util.ArrayList`

because it tried to access `pixelRatio` option at index 6 which is `Double` instead of `size` at index 5 which is `ArrayList<Double>` causing MapboxMapFactory to fail creating the map.


evil159:
Thank you @LjubiTech-Maxko for noticing and fixing this!

Issue #380, opened 2024-01-23
iOS 17 Memory Leak


MeintrupStefan:
## We experience a memory leak on iOS that eventually leads to the app crashing

Once we display a MapWidget, the Ram consumption increases by 100+ MB. 
However, this happens only on iOS and not on Android.

The problem is especially prominent, if we render multiple mapWidgets in a ListView. A screen that contains 6 mapWidgets increased the RAM usage by 300MB. Just a couple of clicks are able to crash the app completely because the RAM usage exceeds 2GB.

## Widget we display:
```dart
    MapWidget(
            key: key,
            resourceOptions: ResourceOptions(
              accessToken:
              "....",
            ),
            cameraOptions: CameraOptions(
              center: Point(coordinates: position + Position(-0.9 * (pow(6.5 - zoom, 2) + 1), 0)).toJson(),
              zoom: zoom,
            ),
            styleUri: "mapbox://styles/.....",
            // Disable taps
            onTapListener: (_) {},
            onMapCreated: (map) => map
              ..scaleBar.updateSettings(ScaleBarSettings(enabled: false))
              ..compass.updateSettings(CompassSettings(enabled: false))
              ..gestures.updateSettings(GesturesSettings(
                quickZoomEnabled: false,
                rotateEnabled: false,
                scrollEnabled: false,
                pitchEnabled: false,
                pinchToZoomEnabled: false,
                doubleTapToZoomInEnabled: false,
                doubleTouchToZoomOutEnabled: false,
                pinchPanEnabled: false,
                pinchToZoomDecelerationEnabled: false,
                rotateDecelerationEnabled: false,
                scrollDecelerationEnabled: false,
                simultaneousRotateAndPinchToZoomEnabled: false,
              ))
              ..logo.updateSettings(LogoSettings(
                marginBottom: 10,
                marginLeft: 10,
              ))
              ..attribution.updateSettings(AttributionSettings(
                position: OrnamentPosition.TOP_RIGHT,
                iconColor: Palette.textSwatch.shade300.value,
                marginTop: 16,
                marginRight: 16,
                clickable: true,
              )),
          );
```

### System details:

Tested on iOS 17.2.1
mapbox_maps_flutter: ^0.5.1
flutter doctor -v:
<details>
[✓] Flutter (Channel stable, 3.16.5, on macOS 14.1.2 23B92 darwin-arm64, locale en-DE)
[...]

PR #383, opened 2024-01-24
[iOS] Free up resources upon map widget disposal

evil159:

Adds tear down for Pigeon-generated channel handlers to reduce memory consumption when the platform FlutterView goes out of scope.

Here's chart of memory consumption opening/closing a screen with map view couple of times:

  | Before | After |
  | ----- | ----- |
  |  |  |

Before the resources from previous map widget were lingering until a new map widget was about to be created(you can see momentary dips in the "Before" chart just before map screen).

https://github.com/mapbox/mapbox-maps-flutter/issues/380

### Pull request checklist:
 - [x] Add a changelog entry.
 - [ ] Not testable.


maios (approved):

sepios87:
The code in the ios/Classes/MapboxMapController.swift file is commented out even though it allows freeing up memory. I tested this implementation on 0.5.1 by adapting the names and this implementation does correctly fix crashes related to memory overuse on iOS. I might have changed the name of releaseMethodChannels to something more descriptive but personally I would approve this pull request !

android/src/main/kotlin/com/mapbox/maps/mapbox_maps/MapboxMapController.kt
```diff
@@ -181,6 +183,10 @@ class MapboxMapController(
         gestureController.removeListeners()
         result.success(null)
       }
+      "platform#releaseMethodChannels" -> {
+        releaseMethodChannels()
```
sepios87, line 187: It would just be necessary to change the name releaseMethodChannels to a more explicit name

example/integration_test/style/style_test.dart
```diff
@@ -283,15 +283,25 @@ void main() {
 
     expect((await style.getStyleLightProperty("flat-light-id", "color")).value,
         listCloseTo(Colors.red.toRGBAList(), 0.0001));
-    expect((await style.getStyleLightProperty("flat-light-id", "color-transition")).value,
+    expect(
```
sepios87, line 286: revert to the old formatting, there is no point in changing this file

ios/Classes/MapboxMapController.swift
```diff
@@ -186,11 +185,30 @@ class MapboxMapController: NSObject, FlutterPlatformView {
...
+        case "platform#releaseMethodChannels":
+            releaseMethodChannels()
+            result(nil)
         default:
             result(FlutterMethodNotImplemented)
         }
     }
 
+    private func releaseMethodChannels() {
+//        channel.setMethodCallHandler(nil)
```
sepios87, line 197: uncomment this

lib/src/location_settings.dart
```diff
@@ -19,16 +19,22 @@ class LocationSettings {
   Future<void> updateSettings(LocationComponentSettings settings) async {
     if (settings.locationPuck == null) {
       // If locationPuck is not set, fallback to use DefaultLocationPuck2D.
-      settings.locationPuck = LocationPuck(locationPuck2D: DefaultLocationPuck2D());
+      settings.locationPuck =
```
sepios87, line 22: revert to the old formatting, there is no point in changing this file

lib/src/mapbox_map.dart
```diff
@@ -144,8 +144,8 @@ class MapboxMap extends ChangeNotifier {
       StyleManager(binaryMessenger: _proxyBinaryMessenger);
 
   /// The interface to set the location puck.
-  late LocationSettings location =
-      LocationSettings(_LocationComponentSettingsInterface(
+  late LocationSettings location = LocationSettings(
```
sepios87, line 147: revert to the old formatting, there is no point in changing this part

lib/src/mapbox_maps_platform.dart
```diff
@@ -137,20 +137,12 @@ class _MapboxMapsPlatform {
...
   }
 
-  void dispose() {
+  void dispose() async {
+    await _channel.invokeMethod('platform#releaseMethodChannels');
     _suffixesRegistry.releaseSuffix(_channelSuffix);
     _channel.setMethodCallHandler(null);
   }
 
-  Future<void> addEventListener(_MapEvent event) async {
```
sepios87, line 145: I do not know if it is useful

lib/src/pigeons/gesture_listeners.dart
```diff
@@ -10,6 +10,7 @@ List<Object?> wrapResponse(
   }
   return <Object?>[error.code, error.message, error.details];
 }
+
```
sepios87, line 13: revert this change

sepios87 (commented):
Please review the comments but otherwise good job fixing memory leaks on iOS I confirm!

Issue #323, opened 2023-12-13
Map freezes/goes blank when backgrounded on Pixel 6


chris-wickens:
After loading the example app in either debug or release mode, I background the app for two minutes, return to it and find the map has either gone blank or frozen entirely. Two minutes is enough time for it to occur consistently, but it can happen as a result of quickly switching to a different app and back. The logs don't show anything that would indicate the cause, at least to my eyes.

Android version 14 AP11.231020.016.A1

![Screenshot_20231213-143236](https://github.com/mapbox/mapbox-maps-flutter/assets/3374071/07f60079-ab6b-4176-9291-17c732d13ed3)

<details>
<summary>Logs</summary>
I/DisplayManager(30476): DLD(CHANGED, display=0, mEventsMask=111, mPackageName=com.mapbox.maps.flutter.example, msg.obj=DisplayInfo{"Built-in Screen", displayId 0, displayGroupId 0, FLAG_SECURE, FLAG_SUPPORTS_PROTECTED_BUFFERS, FLAG_TRUSTED, real 1080 x 2400, largest app 2272 x 2209, smallest app 1080 x 943, appVsyncOff 6233333, presDeadline 11500000, mode 2, renderFrameRate 90.0, defaultMode 2, userPreferredModeId -1, modes [{id=1, width=1080, height=2400, fps=60.0, alternativeRefreshRates=[90.0], supportedHdrTypes=[2, 3, 4]}, {id=2, width=1080, height=2400, fps=90.0, alternativeRefreshRates=[60.0], supportedHdrTypes=[2, 3, 4]}], hdrCapabilities HdrCapabilities{mSupportedHdrTypes=[2, 3, 4], mMaxLuminance=800.0, mMaxAverageLuminance=120.0, mMinLuminance=5.0E-4}, userDisabledHdrTypes [], minimalPostProcessingSupported false, rotation 0, state ON, committedState ON}, listener=class android.view.ViewRootImpl$3)
[...]

Issue #205, opened 2023-07-12
Black map after resume on Android


artemsorochan:
Map becomes black when resumes the app on Android. It doesn't happen on all devices. Sometimes it flickers for 1 second, sometimes stay longer or even forever until app restart. We didn't find any relations to the code, therefore can't provide any reproducable code. Any suggestion would be helpful. We use PointAnnotationManager, PolylineAnnotationManager, PolygonAnnotationManager, OnPointAnnotationClickListener

mapbox_maps_flutter: 0.4.4
 Flutter (Channel stable, 3.10.3)

T-moz:
I also faced this issues.

The problem is also present with https://github.com/maplibre/flutter-maplibre-gl

evil159:
Hi all, this issue should be fixed in https://github.com/mapbox/mapbox-maps-flutter/pull/391, feel free to reopen this ticket if you have any feedback!

PR #391, opened 2024-02-02
Configurable android view hosting mode

evil159:

* Adds a `androidHostingMode` construction parameter to `MapWidget`, allowing to configure the way map view is being hosted in Flutter.
* Changes the default hosting mode from Virtual Display to Hybrid Composition.

Latest Android 14 QPR beta 3 makes the map view to stop rendering on resume from background in Virtual Display mode, switching to Hybrid Composition helps[TODO: Link Android flutter ticket]. Similar issue with the latest update of Android 14 on Samsung Galaxy phones. https://github.com/flutter/flutter/issues/139039 https://github.com/flutter/flutter/issues/139630

Addresses: https://github.com/mapbox/mapbox-maps-flutter/issues/323 https://github.com/mapbox/mapbox-maps-flutter/issues/205
### Pull request checklist:
 - [x] Add a changelog entry.
     Not testable. 

### DIFF
--- a/lib/src/map_widget.dart
+++ b/lib/src/map_widget.dart
@@ -220,8 +220,8 @@ class _MapWidgetState extends State<MapWidget> {
       'mapboxPluginVersion': '1.0.0-beta.2'
     };
 
-    return (_mapboxMapsPlatform.buildView(widget.androidHostingMode,
-        creationParams, onPlatformViewCreated, widget.gestureRecognizers) as PlatformViewLink).createState().;
+    return _mapboxMapsPlatform.buildView(widget.androidHostingMode,
+        creationParams, onPlatformViewCreated, widget.gestureRecognizers);
   }
 
   @override
--- a/lib/src/map_widget.dart
+++ b/lib/src/map_widget.dart
@@ -1,5 +1,33 @@
 part of mapbox_maps_flutter;
 
+/// A mode for platform MapView to be hosted in Flutter on Android platform.
+///
+/// As per https://github.com/flutter/flutter/wiki/Android-Platform-Views#selecting-a-mode
+@experimental
+enum AndroidPlatformViewHostingMode {
+  /// Texture Layer Hybrid Composition with fallback to Virtual Display,
+  /// when the current SDK version is <23 or [MapWidget.textureView] is `false`.
+  ///
+  /// https://github.com/flutter/flutter/wiki/Texture-Layer-Hybrid-Composition
+  TLHC_VD,
+
+  /// Use Texture Layer Hybrid Composition hosting mode with fallback to Hybrid Composition,
+  /// when the current SDK version is <23 or [MapWidget.textureView] is `false`.
+  ///
+  /// https://github.com/flutter/flutter/wiki/Texture-Layer-Hybrid-Composition
+  TLHC_HC,
+
+  /// Always use Hybrid Composition hosting mode.
+  ///
+  /// https://github.com/flutter/flutter/wiki/Hybrid-Composition
+  HC,
+
+  /// Always use Virtual Display hosting mode.
+  ///
+  /// https://github.com/flutter/flutter/wiki/Virtual-Display
+  VD,
+}
+
 /// A MapWidget provides an embeddable map interface.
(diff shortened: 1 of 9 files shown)


yunikkk:
> Latest Android 14 QPR beta 3 makes the map view to stop rendering in Virtual Display mode, switching to Hybrid Composition helps

Nit - stops on resume from background.
I would also linked corresponding flutter repo tickets.

kiryldz:
@evil159 couple questions / proposals:
1. can we link issues in Flutter repo here so that we somehow track when this will be fixed?
2. any known limitations of changing hosting mode? if there are - perhaps it makes sense _not_ change the default value but just introduce the API and let the customers decide themselves based on the device model what hosting mode should be used

lib/src/map_widget.dart
```diff
@@ -186,8 +220,8 @@ class _MapWidgetState extends State<MapWidget> {
       'mapboxPluginVersion': '1.0.0-beta.2'
     };
 
-    return _mapboxMapsPlatform.buildView(
-        creationParams, onPlatformViewCreated, widget.gestureRecognizers);
+    return (_mapboxMapsPlatform.buildView(widget.androidHostingMode,
+        creationParams, onPlatformViewCreated, widget.gestureRecognizers) as PlatformViewLink).createState().;
```
yunikkk, line 224: Can be either `PlatformViewLink` or `AndroidView` - force cast could fail right?

evil159: Thanks for catching this! Remnants from some debugging, removed.

evil159:
> @evil159 couple questions / proposals:
> 
> 1. can we link issues in Flutter repo here so that we somehow track when this will be fixed?
> 2. any known limitations of changing hosting mode? if there are - perhaps it makes sense _not_ change the default value but just introduce the API and let the customers decide themselves based on the device model what hosting mode should be used

1. I've linked the existing tickets(Samsung specific), will link the Android beta one later on
2. Comparing [Virtual Display](https://github.com/flutter/flutter/wiki/Virtual-Display) (VD), [Hybrid Composition](https://github.com/flutter/flutter/wiki/Hybrid-Composition) (HC) and [Texture Layer Hybrid Composition](https://github.com/flutter/flutter/wiki/Texture-Layer-Hybrid-Composition) (TLHC), it seems to me that the priority order is:
* TLHC - newest, combines best of VD and HS while addressing their issues
* HC - predecessor of TLHC, addresses some of VD drawbacks
* VD - the original mode, number of issues caused by it's nature
Of course, each mode has some unique downsides and even VD can shine in right circumstances, but IMO the general order is the above.
I think the ultimate goal for us would be to go with TLHC, but for now HC is confirmed to be working in all conditions, so it is logical to bump it up as well.
I don't think changing hosting mode should be considered a breaking change, so we could change that in the near future should we be compelled to. Besides, those who want to be pinned to a certain mode - are able to do that with the constructor parameter.

kiryldz (approved):
I'd wait for @yunikkk 's expert review, from my high-level overview - LGTM. And - which is more important - fixes the actual issue. BTW perhaps we can add this in changelog as well as bug fix but up to you.

yunikkk (approved):

yunikkk:
@evil159 LGTM!

lib/src/mapbox_maps_platform.dart
```diff
@@ -137,6 +174,27 @@ class _MapboxMapsPlatform {
...
+          AndroidPlatformViewHostingMode hostingMode) {
+    switch (hostingMode) {
+      case AndroidPlatformViewHostingMode.TLHC_VD:
+        return PlatformViewsService.initAndroidView;
+      case AndroidPlatformViewHostingMode.TLHC_HC:
+        return PlatformViewsService.initSurfaceAndroidView;
+      case AndroidPlatformViewHostingMode.HC:
+        return PlatformViewsService.initExpensiveAndroidView;
+      case AndroidPlatformViewHostingMode.VD:
+        throw "Unexpected hostring mode(VD) when selecting an android view controller";
```
pengdev, line 194: nit: typo hostring -> hosting

PR #405, opened 2024-02-10
Apply extended debug environment during android build

evil159:

Applies extended debug environment for Android platform.

Enabling extended debug capabilities.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -9,7 +9,7 @@ buildscript {
     }
 
     dependencies {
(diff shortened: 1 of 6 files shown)


kiryldz (commented):
If it's a public feature (allowing to build from source mapbox-maps-android) - do we want to document it anyhow?

evil159:
> If it's a public feature (allowing to build from source mapbox-maps-android) - do we want to document it anyhow?

This is not available publicly.

kiryldz (approved):

PR #408, opened 2024-02-12
Switch Pigeon to Swift generation

evil159:

Replaces `objc_header_out` and `objc_source_out` with `swift_out` in Pigeon command line arguments. Thus making Pigeon generate Swift code.

Improved compatibility with our Swift middleware layer.


yunikkk (approved):

yunikkk:
Build fails tho @evil159 =\

evil159:
> Build fails tho @evil159 =\

SwiftLint is overzelous with generated files :D Have to exclude them

ios/Classes/CircleAnnotationMessager.swift
```diff
@@ -0,0 +1,601 @@
+// Autogenerated from Pigeon (v17.0.0), do not edit directly.
```
pjleonard37, line 1: There's a type in this file name. Should be CircleAnnotationMessenger

pjleonard37 (approved):

ios/Classes/CircleAnnotationMessager.swift
evil159: Seems to be a bigger change as it will affect iOS, Android and Dart code(through codegen), I've created a ticket to tackle it separately https://mapbox.atlassian.net/browse/MAPSFLT-177

PR #423, opened 2024-02-29
Switch Android Firebase tests to Pixel 8

evil159:

Switch Android Firebase tests to be run on a Pixel 8 device.

After recent outage Pixel 6(our current) devices seem to be still stuck on Firebase.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,7 +18,7 @@ commands:
   setup-gcloud:
     steps:
       - gcp-cli/setup:
-          version: 449.0.0
+          version: 466.0.0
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -36,7 +36,7 @@ jobs:
         default: example/build/app/outputs/apk
     docker:
       - image: cimg/android:2022.09
-    resource_class: xlarge
+    resource_class: 2xlarge
```
evil159, line 39: Bumped this one as well to prevent OOM errors during the build.

pjleonard37 (approved):

Issue #190, opened 2023-06-08
bug! onTapListener gets a ScreenCoordinate which should be in pixel units but its giving in longitude and latitude units instead!


ember11498:
i want to use ScreenBox to queryfeatures i was using onTapListener to know where the user taps on the screen, unfortunately it seems there is a bug and ScreeCoordinates are actual map coordinates instead of pixel coordinates!

evil159:
This should be addressed by https://github.com/mapbox/mapbox-maps-flutter/pull/473

noforeignland:
This is still a problem in April 2024 - is #473 pull request going to be merged in the next realese?

maios:
We believe this issue has been fixed in our recent releases. Please update to the latest version and verify if the problem persists.

Additionally, since version 2.9.0, we’ve introduced the Interactive Maps Element API. We encourage you to use this API to add interaction to a map, layer, or other elements. Documentations can be found [here](https://docs.mapbox.com/flutter/maps/guides/user-interaction/interactions/)

This ticket will now be closed. If you continue to experience any issues, feel free to reopen it. Thank you!

Issue #81, opened 2023-02-10
onLongTapListener returns Lat/Lng not real ScreenCoordinate


LorenzSchueler:
`onLongTapListener` returns a `ScreenCoordinate`, however the `x` and `y` values correspond to the lat/ lng of the position and not the coordinates on the screen (where 0, 0 is top left).

evil159:
@LorenzSchueler Thank you for the report! I can confirm that this is a bug and we have it on our roadmap to fix it. Additionally we are considering reshaping that bit of API to report both screen and geographical coordinates of an interacted point.

PR #473, opened 2024-04-04
Add geographical and screen coordinates to map interaction callbacks

evil159:

Introduces `MapContentGestureContext` carrying touch point and projected map coordinate under the touch point of a gesture.

Changes tap, long tap and move gesture callbacks to use `MapContentGestureContext` instead of `ScreenCoordinate`.

`ScreenCoordinate` is not a correct type to represent geographical position. This is also a matter of ergonomics of providing both touch and geographical positions of a gesture.

Addresses: https://mapbox.atlassian.net/browse/MAPSFLT-111, https://github.com/mapbox/mapbox-maps-flutter/issues/81 and https://github.com/mapbox/mapbox-maps-flutter/issues/190.

### Pull request checklist:
 - [x] Add a changelog entry.


lib/src/pigeons/gesture_listeners.dart
```diff
@@ -4,13 +4,48 @@
...
+    return <Object?>[
+      touchPosition.encode(),
+      point.encode(),
+    ];
+  }
+
+  static MapContentGestureContext decode(Object result) {
+    result as List<Object?>;
+    return MapContentGestureContext(
+      touchPosition: ScreenCoordinate.decode(result[0]! as List<Object?>),
```
pjleonard37, line 30: Is there a way we can avoid the forced `!`?

evil159: This code is Pigeon-generated, it seems that the assumptions here, since the decoding and encoding is both done by Pigeon, they deem to be acceptable to force unwrap it.

pjleonard37: Seems fair 👍

CHANGELOG.md
```diff
@@ -1,7 +1,113 @@
 ### main
 
+* Mark `MapboxMapsOptions.get/setWorldview()` and `MapboxMapsOptions.get/setLanguage()` as experimental.
+* Bump Pigeon to 17.1.2
+* [iOS] Fix crash in `onStyleImageMissingListener`.
+* Deprecate `cameraForCoordinates`, please use `cameraForCoordinatesPadding` instead.
+* Add a way to disable default puck' image(s) when using `DefaultLocationPuck2D` by passing an empty byte array, for example, the following code shows a puck 2D with custom top image, default bearing image and no shadow image.
```
pjleonard37, line 7:
```suggestion
* Add a way to disable default puck's image(s) when using `DefaultLocationPuck2D`. By passing an empty byte array, for example, the following code shows a puck 2D with custom top image, default bearing image and no shadow image.
```

CHANGELOG.md
```diff
@@ -1,7 +1,113 @@
...
+onTapListener: { (context)
+    final coordinates = context.point.coordinates; // Position
+    final touchPosition = context.touchPosition; // ScreenCoordinate
+    ...
+}
+```
+
+
+* Bump Pigeon to 17.1.2
+* [iOS] Fix crash in `onStyleImageMissingListener`.
```
pjleonard37, line 64: These are duplicated above. 
```suggestion
```

CHANGELOG.md
```diff
@@ -1,7 +1,113 @@
...
+* Fix camera center not applied from map init options.
 * [iOS] Free up resources upon map widget disposal. This should help to reduce the amount of used memory when previously shown map widget is removed from the widget tree.
 
+### 1.0.0
+
+* Add `MapboxMapsOptions.get/setWorldview()` and ``MapboxMapsOptions.get/setLanguage()`. Use this to to adjust administrative boundaries/map language based on the map's audience.
```
pjleonard37, line 70:
```suggestion
* Add `MapboxMapsOptions.get/setWorldview()` and `MapboxMapsOptions.get/setLanguage()`. Use this to to adjust administrative boundaries/map language based on the map's audience.
```

pjleonard37 (approved):
Looks really nice 😎

Issue #477, opened 2024-04-08
1.1.0-rc.1  crashes on Android


autolainen:
I am trying to setup just a basic map view and when a route with map open, all the app crashes with stacktrace:

E/ThemeUtils(21728): View class com.mapbox.maps.plugin.compass.CompassViewImpl is an AppCompat widget that can only be used with a Theme.AppCompat theme (or descendant).
E/ThemeUtils(21728): View class com.mapbox.maps.plugin.logo.LogoViewImpl is an AppCompat widget that can only be used with a Theme.AppCompat theme (or descendant).
E/ThemeUtils(21728): View class com.mapbox.maps.plugin.attribution.AttributionViewImpl is an AppCompat widget that can only be used with a Theme.AppCompat theme (or descendant).
I/PlatformViewsController(21728): Using hybrid composition for platform view: 0
D/TrafficStats(21728): tagSocket(239) with statsTag=0xffffffff, statsUid=-1
D/TrafficStats(21728): tagSocket(220) with statsTag=0xffffffff, statsUid=-1
D/CompatibilityChangeReporter(21728): Compat change id reported: 236825255; UID 10275; state: DISABLED
E/flutter (21728): [ERROR:flutter/fml/platform/android/jni_util.cc(206)] java.lang.IllegalStateException: Please ensure that the hosting activity/fragment is a valid LifecycleOwner
E/flutter (21728): 	at com.mapbox.maps.plugin.lifecycle.ViewLifecycleOwner.doOnAttached(ViewLifecycleOwner.kt:70)
E/flutter (21728): 	at com.mapbox.maps.plugin.lifecycle.ViewLifecycleOwner.<init>(ViewLifecycleOwner.kt:60)
E/flutter (21728): 	at com.mapbox.maps.plugin.lifecycle.MapboxLifecyclePluginImpl.registerLifecycleObserver(MapboxLifecyclePluginImpl.kt:30)
E/flutter (21728): 	at com.mapbox.maps.plugin.MapPluginRegistry.onAttachedToWindow(MapPluginRegistry.kt:192)
E/flutter (21728): 	at com.mapbox.maps.MapController.onAttachedToWindow$sdk_release(MapController.kt:366)
E/flutter (21728): 	at com.mapbox.maps.MapView.onAttachedToWindow(MapView.kt:135)
E/flutter (21728): 	at android.view.View.dispatchAttachedToWindow(View.java:23198)
[...]

Issue #308, opened 2023-11-26
E/flutter (12658): [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: PlatformException(error, java.lang.IllegalStateException: Please ensure that the hosting activity/fragment is a valid LifecycleOwner


OnlineNomad:
help

OnlineNomad:
pubsec.yaml:
name: aqari_official
description: "flutter official mapbox integration"
# The following line prevents the package from being accidentally published to
# pub.dev using `flutter pub publish`. This is preferred for private packages.
publish_to: "none" # Remove this line if you wish to publish to pub.dev

# The following defines the version and build number for your application.
# A version number is three numbers separated by dots, like 1.2.43
# followed by an optional build number separated by a +.
# Both the version and the builder number may be overridden in flutter
# build by specifying --build-name and --build-number, respectively.
# In Android, build-name is used as versionName while build-number used as versionCode.
# Read more about Android versioning at https://developer.android.com/studio/publish/versioning
# In iOS, build-name is used as CFBundleShortVersionString while build-number is used as CFBundleVersion.
# Read more about iOS versioning at
# https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html
# In Windows, build-name is used as the major, minor, and patch parts
# of the product and file versions while build-number is used as the build suffix.
version: 1.0.0+1

environment:
  sdk: ">=3.2.0 <4.0.0"

# Dependencies specify other packages that your package needs in order to work.
# To automatically upgrade your package dependencies to the latest versions
# consider running `flutter pub upgrade --major-versions`. Alternatively,
# dependencies can be manually updated by changing the version numbers below to
# the latest version available on pub.dev. To see which dependencies have newer
# versions available, run `flutter pub outdated`.
dependencies:
  flutter:
    sdk: flutter

  # The following adds the Cupertino Icons font to your application.
  # Use with the CupertinoIcons class for iOS style icons.
  cupertino_icons: ^1.0.2
  mapbox_maps_flutter: ^0.5.0
dev_dependencies:
  flutter_test:
    sdk: flutter

  # The "flutter_lints" package below contains a set of recommended lints to
  # encourage good coding practices. The lint set provided by the package is
  # activated in the `analysis_options.yaml` file located at the root of your
  # package. See that file for information about deactivating specific lint
  # rules and activating additional ones.
[...]

Issue #291, opened 2023-11-02
Adding MapWidget causes Unhandled Exception


kinga112:
I just added the widget as a child to a container to try to get it working.

MapWidget(
            resourceOptions: ResourceOptions(
                accessToken: dotenv.env['MAPBOX_ACCESS_TOKEN']!))

I added my access token to the locations posted in the read me.

E/flutter (32061): [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: PlatformException(error, java.lang.IllegalStateException: Please ensure that the hosting activity/fragment 
is a valid LifecycleOwner
E/flutter (32061):      at com.mapbox.maps.plugin.lifecycle.ViewLifecycleOwner.doOnAttached(ViewLifecycleOwner.kt:68)

Ashok-Media:
got any solution? @kinga112

evil159:
Hi all, this issue should be fixed in https://github.com/mapbox/mapbox-maps-flutter/pull/478, feel free to reopen this ticket if you have any feedback!

Issue #259, opened 2023-10-05
IllegalStateException: Please ensure that the hosting activity/fragment is a valid LifecycleOwner


amias-samir:
```
> Expecting binder but got null!

>   E  [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: PlatformException(error, java.lang.IllegalStateException: Please ensure that the hosting activity/fragment is a valid LifecycleOwner
>                                                                 	at com.mapbox.maps.plugin.lifecycle.ViewLifecycleOwner.doOnAttached(ViewLifecycleOwner.kt:68)
>                                                                 	at com.mapbox.maps.plugin.lifecycle.ViewLifecycleOwner.<init>(ViewLifecycleOwner.kt:58)
>                                                                 	at com.mapbox.maps.plugin.lifecycle.MapboxLifecyclePluginImpl.registerLifecycleObserver(MapboxLifecyclePluginImpl.kt:30)
>                                                                 	at com.mapbox.maps.plugin.MapPluginRegistry.onAttachedToWindow(MapPluginRegistry.kt:173)
>                                                                 	at com.mapbox.maps.MapController.onAttachedToWindow$sdk_publicRelease(MapController.kt:346)
>                                                                 	at com.mapbox.maps.MapView.onAttachedToWindow(MapView.kt:129)
>                                                                 	at android.view.View.dispatchAttachedToWindow(View.java:21511)
>                                                                 	at android.view.ViewGroup.dispatchAttachedToWindow(ViewGroup.java:3590)
>                                                                 	at android.view.ViewGroup.dispatchAttachedToWindow(ViewGroup.java:3597)
>                                                                 	at android.view.ViewGroup.addViewInner(ViewGroup.java:5408)
>                                                                 	at android.view.ViewGroup.addView(ViewGroup.java:5188)
>                                                                 	at android.view.ViewGroup.addView(ViewGroup.java:5128)
>                                                                 	at android.view.ViewGroup.addView(ViewGroup.java:5100)
>                                                                 	at io.flutter.plugin.platform.PlatformViewsController.configureForTextureLayerComposition(PlatformViewsController.java:648)
>                                                                 	at io.flutter.plugin.platform.PlatformViewsController.access$900(PlatformViewsController.java:49)
>                                                                 	at io.flutter.plugin.platform.PlatformViewsController$1.createForTextureLayer(PlatformViewsController.java:225)
[...]

PR #478, opened 2024-04-08
Fix crash due to lifecycle plugin.

yunikkk:

Addresses the issue that was caused by the `maps-lifecycle` plugin being pulled to the build.
Normally it was excluded from the flutter plugin explicitly [here](https://github.com/mapbox/mapbox-maps-flutter/blob/main/android/build.gradle#L66), and instead lifecycle methods onStart/onStop/onDestroy were called directly on the MapView.

However, other 3-party plugins could still bring the dependency. 
The plugin then crashes internally within [ViewLifecycleOwner](https://github.com/mapbox/mapbox-maps-android/blob/main/plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/ViewLifecycleOwner.kt#L69).

In this PR custom Lifecycle is implemented to manage MapView lifecycle calls, `maps-lifecycle` dependency is no longer excluded. 
Custom Lifecycle mirrors lifecycle of the hosting Flutter activity + additionally triggers MapView stop/destroy when widget is disposed but Flutter activity keeps living (that happens easily when MapView is added / removed from the screen).

Should fix multiple reported issues about `java.lang.IllegalStateException: Please ensure that the hosting activity/fragment is a valid LifecycleOwner` : #259, #152, #60, #291, #477, #308, #50. 

### Pull request checklist:
 - [x] Add a changelog entry.

### DIFF
--- a/android/src/main/kotlin/com/mapbox/maps/mapbox_maps/MapboxMapsPlugin.kt
+++ b/android/src/main/kotlin/com/mapbox/maps/mapbox_maps/MapboxMapsPlugin.kt
@@ -13,21 +13,12 @@ import io.flutter.plugin.common.MethodChannel.MethodCallHandler
 import io.flutter.plugin.common.MethodChannel.Result
 
 /** MapboxMapsPlugin */
-class MapboxMapsPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
+class MapboxMapsPlugin : FlutterPlugin, ActivityAware {
   private val optionsController = MapboxOptionsController()
 
   private var lifecycle: Lifecycle? = null
 
-  // / The MethodChannel that will the communication between Flutter and native Android
-  // /
-  // / This local reference serves to register the plugin with the Flutter Engine and unregister it
-  // / when the Flutter Engine is detached from the Activity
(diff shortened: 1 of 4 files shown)


android/src/main/kotlin/com/mapbox/maps/mapbox_maps/MapboxMapsPlugin.kt
```diff
@@ -13,21 +13,12 @@ import io.flutter.plugin.common.MethodChannel.MethodCallHandler
...
+class MapboxMapsPlugin : FlutterPlugin, ActivityAware {
   private val optionsController = MapboxOptionsController()
 
   private var lifecycle: Lifecycle? = null
 
-  // / The MethodChannel that will the communication between Flutter and native Android
-  // /
-  // / This local reference serves to register the plugin with the Flutter Engine and unregister it
-  // / when the Flutter Engine is detached from the Activity
-  private lateinit var channel: MethodChannel
```
yunikkk, line 25: Some outdated channel unused channel - got rid of it

evil159 (approved):
🔥

PR #487, opened 2024-04-12
Bump plugin version to 1.1.0

evil159:

Updates plugin version to 1.1.0.

Preparation for the release.

### Pull request checklist:
 - [x] Add a changelog entry.


maios:
The changelog does not look very good imo, there are duplicated entries here and there between main and 1.1.0 and 1.1.0-rc.1 section, could we refine it?

evil159:
> The changelog does not look very good imo, there are duplicated entries here and there between main and 1.1.0 and 1.1.0-rc.1 section, could we refine it?

IIRC this is how we usually prepare release notes - beta and rc notes are included into the stable release, so those who follow stable releases don't have to hunt the notes down across pre-releases.

pjleonard37 (approved):

PR #513, opened 2024-04-26
MapWidget snapshotting

evil159:

Adds `MapWidget.snapshot()` method allowing to capture a snapshot image of the current state of the map widget.

API parity with platform SDKs. 

### Pull request checklist:
 - [x] Add a changelog entry.

### DIFF
--- a/android/src/main/kotlin/com/mapbox/maps/mapbox_maps/MapboxMapController.kt
+++ b/android/src/main/kotlin/com/mapbox/maps/mapbox_maps/MapboxMapController.kt
@@ -1,6 +1,7 @@
 package com.mapbox.maps.mapbox_maps
 
 import android.content.Context
+import android.graphics.Bitmap
 import android.view.View
 import androidx.lifecycle.DefaultLifecycleObserver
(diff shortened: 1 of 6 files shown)


pjleonard37 (approved):

android/src/main/kotlin/com/mapbox/maps/mapbox_maps/MapboxMapController.kt
```diff
@@ -202,6 +204,21 @@ class MapboxMapController(
         dispose()
         result.success(null)
       }
+      "map#snapshot" -> {
+        val mapView = mapView
+        val snapshot = mapView?.snapshot()
+        if (mapView == null) {
+          result.error("2342345", "Failed to create snapshot: map view is not found.", null)
```
maios, line 211: what is the meaning of this error code? do we have a list of error codes somewhere we can look up?

evil159: I wasn't sure what to put there, so just randomly mashed my fingers against the number row 🙈

Issue #500, opened 2024-04-19
IconAnchor enum is not encoded/decoded correctly


bf-vs:
`IconAnchor.BOTTOM_RIGHT` is endoced as 'bottom_right' where mapbox expects 'bottom-right' (- vs _) therefore it does not work.

This line should be fixed https://github.com/mapbox/mapbox-maps-flutter/blob/1daef17c48fd927db442a5d19f7aca06a1ffdc63/lib/src/style/layer/symbol_layer.dart#L277

evil159:
Hi @bf-vs, thank you for reporting this issue, it is addressed in https://github.com/mapbox/mapbox-maps-flutter/pull/523.

O-Hannonen:
Related to #455 and #161.

Closing this as fixed by #523, the fix in Flutter side landed in version 2.0.0. Thanks for reporting!

Issue #455, opened 2024-03-20
IconAnchor


FronnieSW:
Hello, 
currently the icon anchor for bottom_right and bottom_left does not work. I suppose the problem is gonna be in every case, when u have to pass an enum item, that contains ' _ ' , even in the other classes and cases.
The problem is that currently you are doing  : iconAnchor?.toString().split('.').last.toLowerCase(); - In file symbol_layer.dart, which returns ("bottom_left") but for bottom left, you are suppose to pass value ("bottom-left") so when I add .replaceAll("_", "-") to the function, which returns "bottom-left", it works fine. 
Have a nice day.

O-Hannonen:
Related to #500 and #161 

Two separate bugs got mixed into this thread, so clearing it up:
1. The original report (`IconAnchor` on a style layer, e.g. `SymbolLayer`) was fixed
   in mapbox_maps_flutter 2.0.0
2. The crash reported afterwards (`IconAnchor.valueOf` does not support
   `[BOTTOM-LEFT]` when using `PointAnnotationOptions`) was a different bug, in the
   native Android Maps SDK's annotation plugin, not in this package. It was
   fixed in Maps SDK 11.14.0 and shipped in mapbox_maps_flutter 2.10.0

Closing this now as fixed, thanks for reporting and the discussion!

Issue #161, opened 2023-05-11
TextAnchor issue


penhorwood:
I am getting the following error when running 0.4.3 in debug mode when trying to add a point annotation to the map.  My code uses
```
PointAnnotationOptions(
...
    textAnchor: TextAnchor.BOTTOM_LEFT
)
```

```
Exception has occurred.
PlatformException (PlatformException(IllegalArgumentException, java.lang.IllegalArgumentException: No enum constant com.mapbox.maps.extension.style.layers.properties.generated.TextAnchor.BOTTOM-LEFT, Cause: null, Stacktrace: java.lang.IllegalArgumentException: No enum constant com.mapbox.maps.extension.style.layers.properties.generated.TextAnchor.BOTTOM-LEFT
	at java.lang.Enum.valueOf(Enum.java:257)
	at com.mapbox.maps.extension.style.layers.properties.generated.TextAnchor.valueOf(Unknown Source:2)
	at com.mapbox.maps.plugin.annotation.generated.PointAnnotation.getTextAnchor(PointAnnotation.kt:288)
	at com.mapbox.maps.mapbox_maps.annotation.PointAnnotationControllerKt.toFLTPointAnnotation(PointAnnotationController.kt:833)
	at com.mapbox.maps.mapbox_maps.annotation.PointAnnotationController.create(PointAnnotationController.kt:33)
	at com.mapbox.maps.pigeons.FLTPointAnnotationMessager$_PointAnnotationMessager$-CC.lambda$setup$0(FLTPointAnnotationMessager.java:1283)
	at com.mapbox.maps.pigeons.FLTPointAnnotationMessager$_PointAnnotationMessager$$ExternalSyntheticLambda0.onMessage(Unknown Source:2)
```

penhorwood:
PS D:\android\apps\caminha> flutter doctor -v
[√] Flutter (Channel stable, 3.10.6, on Microsoft Windows [Version 10.0.19045.3208], locale en-US)
    • Flutter version 3.10.6 on channel stable at D:\development\flutter-sdk
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision f468f3366c (9 days ago), 2023-07-12 15:19:05 -0700
    • Engine revision cdbeda788a
    • Dart version 3.0.6
    • DevTools version 2.23.1

[√] Windows Version (Installed version of Windows is version 10 or higher)

[√] Android toolchain - develop for Android devices (Android SDK version 33.0.2)
[...]

PR #523, opened 2024-04-30
Fix multiword enum cases encoding and decoding

evil159:

Multi-word enum cases were not encoded/decoded properly - instead of the correct form e.g. `some-value` it would be `some_value` which would cause a PlatformException.

https://github.com/mapbox/mapbox-maps-flutter/issues/500, https://github.com/mapbox/mapbox-maps-flutter/issues/455, https://github.com/mapbox/mapbox-maps-flutter/issues/161, https://mapbox.atlassian.net/browse/MAPSFLT-206

### Pull request checklist:
 - [x] Add a changelog entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,7 @@ onTapListener: { (context)
 
 * Fix camera center not applied from map init options.
 * [iOS] Free up resources upon map widget disposal. This should help to reduce the amount of used memory when previously shown map widget is removed from the widget tree.
+* Fix multi-word enum cases decoding/encoding when being sent to/from the platform side.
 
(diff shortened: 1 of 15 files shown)


bf-vs:
LGTM but why not use [enhanced enums](https://dart.dev/language/enums#declaring-enhanced-enums)? It would be a cleaner solution.

evil159:
> LGTM but why not use [enhanced enums](https://dart.dev/language/enums#declaring-enhanced-enums)? It would be a cleaner solution.

Our code generation tool - Pigeon, doesn't support them.

pjleonard37 (approved):
👍

PR #530, opened 2024-05-03
Bump MapboxMaps SDK to v11.4.0-beta.2

maios:


pjleonard37 (approved):
The changes look good. However, I think you need to run `flutter run` to rebuild the xcodeproj and push those changes so the iOS build tests pass.

PR #579, opened 2024-06-12
Fix Transparent background on iOS

### LINKED ISSUE #415: Transparent background for the map
Is it possible to make the background of the map transparent?

I tried to set style with '{"version": 8, "sources": {}, "layers":[]}' and it works fine on Android devices, but iOS shows black screen instead of transparent one. 

Please advise it there is any way to make it for iOS also.

tranhuongk:
Fix #415 

set `mapView.isOpaque = false`

### Pull request checklist:
 - [ ] Add a changelog entry.

### DIFF
--- a/ios/Classes/MapboxMapController.swift
+++ b/ios/Classes/MapboxMapController.swift
@@ -57,6 +57,7 @@ final class MapboxMapController: NSObject, FlutterPlatformView {
             .set(key: "com.mapbox.common.telemetry.internal.custom_user_agent_fragment", value: "FlutterPlugin/\(pluginVersion)")
 
         mapView = MapView(frame: frame, mapInitOptions: mapInitOptions)
+        mapView.isOpaque = false
         mapboxMap = mapView.mapboxMap
 
         self.registrar = registrar


evil159 (changes requested):
Hi @tranhuongk, thank you for your contribution! Could you make this behavior configurable, plus this option is also applicable for Android, there opacity can be achieved by using a texture view rendering - `MapInitOptions.textureView`.

tranhuongk:
@evil159  with `textureView` 

```
Flag indicating to use a TextureView as render surface for the MapWidget. Only works for Android
```

tranhuongk:
@evil159 @OdNairy Can we get back to this issue?

tranhuongk:
@evil159

m4ta1:
Any news on this issue? Thanks a lot for fixing this!

evil159:
Hi all, sorry for the delayed reply, would it be possible to make this configurable from the outside? Like to have this as init parameter for `MapWidget`?

vishnupanchal3909:
Hi Team,

I am unable to access the isOpaque property. Could you confirm if this property allows making the background of MapWidget transparent?

Currently, I am facing an issue where disabling the atmosphere in globe projection results in a black background outside the Earth. I need to make this area transparent or set a custom color.

Please let me know ASAP, as I am unable to resolve this issue.

m4ta1:
> Hi all, sorry for the delayed reply, would it be possible to make this configurable from the outside? Like to have this as init parameter for `MapWidget`?

@evil159 thanks for getting back to this issue. Having an init parameter like "background-opacity" for `MapWidget` would be a great solution. 
Situation right now: Working on layer level by setStyleLayerProperty('background', 'background-opacity', 0) makes the background successfully transparent for Android, but has no effect on iOS.

PR #593, opened 2024-06-20
[ios] Fix flaky GeoJsonSource tests

maios:
- Use StyleManager/addSource(_:dataId:) so GeoJsonSource's data will be applied in a background queue.
- Add a delay in GeoJsonSource test when checking for volatile properties, since these properties are added later after the source is added to the style.

**TODO** It might be better for us to adopt StyleDSL from both platforms to Flutter.  I created an [internal ticket](https://mapbox.atlassian.net/browse/MAPSFLT-221) to track work

### Pull request checklist:
 - [x] Add a changelog entry.


pjleonard37:
Closing as this issue was resolved here: https://github.com/mapbox/mapbox-maps-flutter/pull/687/files#diff-e0c8ed8f9a2baf7650e9c91e39a14b023a9143f48a7c9552f51872bbeda52b92L313

PR #608, opened 2024-07-04
Get Flutter asset path in application bundle

maios:

Look up for the path of Flutter assets in application's bundle and pass this value to core renderer for `ModelLayer`
[Sharing assets with the underlying platforms](https://docs.flutter.dev/ui/assets/assets-and-images#sharing-assets-with-the-underlying-platform)

To specify a local asset, you must use `asset://` protocol in the uri.

Flutter assets can be accessed from platform's code, however, they are not just simply looked up by the given path, we first need to get the lookup key from the application's bundle, then use this key to get the correct path and our renderer can use.

```
mapboxMap?.location.updateSettings(LocationComponentSettings(
  locationPuck: LocationPuck(
    locationPuck3D: LocationPuck3D(
      modelUri: "asset://assets/sportcar.glb", // this asset is located at path "~/Frameworks/App.framework/assets/sportcar.glb" for iOS
      modelScale: [_puckScale, _puckScale, _puckScale]))));
```

### Pull request checklist:
 - [x] Add a changelog entry.


android/src/main/kotlin/com/mapbox/maps/mapbox_maps/MapboxOptionsController.kt
```diff
@@ -39,6 +42,10 @@ class MapboxOptionsController : _MapboxMapsOptions, _MapboxOptions {
     return ""
   }
 
+  override fun getFlutterAssetPath(flutterAssetUri: String?): String? {
+    return flutterAssetUri?.replace("""^asset://(.+)""".toRegex(), "asset://${flutterAssets.getAssetFilePathBySubpath("$1")}")
```
OdNairy, line 46: Do we actually need regexp here? Can we replace a `asset://` prefix with the full path instead?

OdNairy, line 46: Any chance we would break existing absolute path references? Was it possible for customers to provide full assets path themselves?

OdNairy (approved):

android/src/main/kotlin/com/mapbox/maps/mapbox_maps/MapboxOptionsController.kt
maios: > The returned file path is relative to the Android app's standard assets directory. Therefore, the returned path is appropriate to pass to Android's AssetManager, but the path is not appropriate to load as an absolute path.

Since the path is not an absolute path, I have to keep `asset` scheme here so core renderer would understand it

OdNairy: I meant to replace `asset://` with `asset:///var/mobile/full/path` way.

maios: clients can always provide an absolute path to the asset, however, we'd expect them to not provide them with `asset` scheme in the URL. 
This is done similarly to platform SDKs, is to provide an easy way for customers to refer to a local asset, without having to implement their own logic, or use another 3rd party plugin to get the absolute path.

maios: I do not know how to obtain the absolute path of the asset in Android 😅

evil159 (dismissed):

evil159 (approved):

PR #611, opened 2024-07-09
Fix lifecycle crash

evil159:

This PR fixes map view crashing upon host activity destruction when using a cached Flutter engine.

Bugfix.

### Pull request checklist:
 - [x] Add a changelog entry.


maios (dismissed):

android/src/main/kotlin/com/mapbox/maps/mapbox_maps/MapboxMapController.kt
```diff
@@ -104,19 +106,18 @@ class MapboxMapController(
...
+      }
     }
 
     fun dispose() {
       parentLifecycle.removeObserver(this)
-      // fires MapView.onStop
-      lifecycleRegistry.currentState = Lifecycle.State.CREATED
-      // fires MapView.onDestroy
-      lifecycleRegistry.currentState = Lifecycle.State.DESTROYED
     }
```
yunikkk, line 117: Hmm, so now when dispose is called in `onFlutterViewDetached` - lifecycle registry is not reset to CREATED state at all? How does this work, `onFlutterViewDetached` is always called after `onDestroy` ?

evil159: Good point, I couldn't find a confirmation that disposal would always be preceded by onDestroy, Added it back.

yunikkk (approved):

Issue #605, opened 2024-07-04
onLongTapListener on iOS not working as expected


RAglx:
See `GesturesPage` in your examples.

The long press / `onLongTapListener` on Android triggers after a given time when
the finger taps the screen. This is fine.

On iOS it only triggers after lifting the finger, which is not what to expect.

IMHO it should work like the `GestureDetector` functions `onLongPress` / `onLongPressStart` and
not like `onLongPressEnd`.

I'm using MapBox 2.0.0 and:
[✓] Flutter (Channel stable, 3.19.6, on macOS 14.5 23F79 darwin-arm64, locale de-CH)
[✓] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
[✓] Xcode - develop for iOS and macOS (Xcode 15.4)
[✓] Android Studio (version 2023.1)
[✓] VS Code (version 1.90.2)

Flutter 3.19.6 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 54e66469a9 (3 months ago) • 2024-04-17 13:08:03 -0700
Engine • revision c4cd48e186
Tools • Dart 3.3.4 • DevTools 2.31.1

PR #620, opened 2024-07-16
[ios] Adapt GesturesManager/onMapTap and onMapLongPress

maios:
We should take advantage of existing gesture recognizers of MapView, namely onMapTap and onMapLongPress.

Remove extra tap and long press recognizer in GesturesController, replace them with `onMapTap` and `onMapLongPress`
This should fix https://github.com/mapbox/mapbox-maps-flutter/issues/605 as `onMapLongPress` will emit events when the long press is recognized (after a minimum press duration), current approach with target/action will report event when the fingers are lifted off the screen.

### Pull request checklist:
 - [ ] Add a changelog entry.

### DIFF
--- a/ios/Classes/Extensions.swift
+++ b/ios/Classes/Extensions.swift
@@ -374,6 +374,14 @@ extension CGPoint {
         return ScreenCoordinate(x: x, y: y)
     }
 }
+
+extension MapboxMaps.MapContentGestureContext {
+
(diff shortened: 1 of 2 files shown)


evil159 (approved):
Nice 👍

RAglx:
Great. Thank you. 👍

PR #656, opened 2024-08-19
Prepare v2.2.0 release (#655)

maios:
* bump deps to v11.6.0

* bump version to v2.2.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,15 +2,8 @@
 
 * Bump Maps SDK to 11.6.0
 * Update Pigeon to `21.1.0`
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+### 2.2.0
```
evil159, line 1: Usually we combine beta, rc and stable release notes into one it in the Github release, for Flutter I think it would make sense to combine them in the changelog file as well, since it will be published to pub.dev as is.

evil159 (approved):

PR #682, opened 2024-09-09
Use default TileStore for offline example

maios:


evil159 (approved):

baleboy (approved):
Tested on Android device where this was previously failing, works now.

PR #687, opened 2024-09-13
[MAPSFLT-236] Expose Partial GeoJSON updates API 

pjleonard37:

This PR exposes 3 new methods which allow you to add/update/remove GeoJSONSourceFeatures. When you add identifiers associated with each feature, you can change them at runtime rather than needing to set a whole new GeoJSON object. This is especially beneficial for ``GeoJSONSource``s which host a large amount of features - in this case adding a feature can be up to 4x faster with the partial GeoJSON update API.

```dart
mapboxMap.style.addGeoJSONSourceFeatures(sourceId, dataId, features)
mapboxMap.style.updateGeoJSONSourceFeatures(sourceId, dataId, features)
mapboxMap.style.removeGeoJSONSourceFeatures(sourceId, dataId, featureIds)
```

This PR additionally updates the Draw a GeoJSON Line example to show how to use these new methods. Further, it updates `addSource` method to avoid a race condition. 

https://github.com/user-attachments/assets/6badbc71-1eff-4d4b-9021-a68ee6cfd403

Allow developers to update GeoJSON features at runtime. 

### Pull request checklist:
 - [ ] Add a changelog entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,9 @@ Instead of setting a whole new GeoJSON object anew every time a single feature h
 If your features have associated identifiers - you can add, update, and remove them on individual basis in your ``GeoJSONSource``. This is especially beneficial for ``GeoJSONSource``s hosting a large amount of features - in this case adding a feature can be up to 4x faster with the partial GeoJSON update API.
 
 ```dart
-maboxMap.style.addGeoJSONSourceFeatures(sourceId, dataId, features)
-maboxMap.style.updateGeoJSONSourceFeatures(sourceId, dataId, features)
-maboxMap.style.removeGeoJSONSourceFeatures(sourceId, dataId, featureIds)
+mapboxMap.style.addGeoJSONSourceFeatures(sourceId, dataId, features)
+mapboxMap.style.updateGeoJSONSourceFeatures(sourceId, dataId, features)
+mapboxMap.style.removeGeoJSONSourceFeatures(sourceId, dataId, featureIds)
 ```
 
 ### 2.3.0-beta.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,16 @@
+### main
+
+* Add support for partial GeoJSON updates. 
+
+Instead of setting a whole new GeoJSON object anew every time a single feature has changed, now you can apply more granular, partial GeoJSON updates.
+If your features have associated identifiers - you can add, update, and remove them on individual basis in your ``GeoJSONSource``. This is especially beneficial for ``GeoJSONSource``s hosting a large amount of features - in this case adding a feature can be up to 4x faster with the partial GeoJSON update API.
+
+```dart
+maboxMap.style.addGeoJSONSourceFeatures(sourceId, dataId, features)
+maboxMap.style.updateGeoJSONSourceFeatures(sourceId, dataId, features)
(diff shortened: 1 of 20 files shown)


lib/src/style/style.dart
```diff
@@ -438,3 +452,7 @@ extension StyleColorList on List {
     }
   }
 }
+
+extension GeoJSONSourceInternal on GeoJsonSource {
+  String? get internalData => '$_data';
```
evil159, line 457: Why this wrapper is needed?

CHANGELOG.md
```diff
@@ -1,3 +1,16 @@
...
+
+* Add support for partial GeoJSON updates. 
+
+Instead of setting a whole new GeoJSON object anew every time a single feature has changed, now you can apply more granular, partial GeoJSON updates.
+If your features have associated identifiers - you can add, update, and remove them on individual basis in your ``GeoJSONSource``. This is especially beneficial for ``GeoJSONSource``s hosting a large amount of features - in this case adding a feature can be up to 4x faster with the partial GeoJSON update API.
+
+```dart
+maboxMap.style.addGeoJSONSourceFeatures(sourceId, dataId, features)
+maboxMap.style.updateGeoJSONSourceFeatures(sourceId, dataId, features)
+maboxMap.style.removeGeoJSONSourceFeatures(sourceId, dataId, featureIds)
```
evil159, line 11:
```suggestion
mapboxMap.style.addGeoJSONSourceFeatures(sourceId, dataId, features)
mapboxMap.style.updateGeoJSONSourceFeatures(sourceId, dataId, features)
mapboxMap.style.removeGeoJSONSourceFeatures(sourceId, dataId, featureIds)
```

evil159 (dismissed):
Nice that you solved the mystery with volatile properties 👍

evil159:
@pjleonard37 Could you update the templates as well?

lib/src/style/style.dart
pjleonard37: It was the only way to get access to the internal `_data` property as the [getter](https://github.com/mapbox/mapbox-maps-flutter/blob/main/lib/src/style/source/geojson_source.dart#L47) returns the value from the style.

CHANGELOG.md
pjleonard37: Good catch 😅

lib/src/style/style.dart
evil159: @pjleonard37 Shouldn't `_data` be accessible from inside the package?

pjleonard37: Updated this in the latest commit after live discussion. 👍

pjleonard37:
> @pjleonard37 Could you update the templates as well?
Yes, I updated them locally and will push a PR once this lands

evil159 (approved):

wwyung:
@pjleonard37, thanks for making these API available, great job! However when I tried to add some new feature source to an existing source layer using addGeoJSONSourceFeatures(), I am not seeing the feature showing up on the map. I did see onSourceDataLoadedListener callback invoked successfully, and there is no call to onMapLoadErrorListener callback. How do I go about to debug and root cause the issue?

pjleonard37:
Hi @wwyung -- 

Thanks for reporting this. Could you open a new issue in this repo with a simple test case that reproduces the issue? We can then use that to investigate on our side.

wwyung:
@pjleonard37, nevermind my request, it was found as a bug in my code. Thanks for the response.

PR #689, opened 2024-09-17
Debug logging

evil159:

This PR adds proxying Mapbox Maps SDK logs from platform to flutter side in debug configuration to print them in debug console with `log()` function from `dart:developer` package.

Logs are proxied **only** in debug configuration and can be disabled completely by passing environment flag `MAPBOX_LOG_DEBUG` with `false` value.

Easier debugging experience.

### Pull request checklist:
 - [x] Add a changelog entry.


lib/src/mapbox_maps_options.dart
```diff
@@ -4,7 +4,9 @@ part of mapbox_maps_flutter;
 final class MapboxOptions {
   static _MapboxOptions _options = _MapboxOptions();
 
-  MapboxOptions._() {}
+  MapboxOptions._() {
+    LogConfiguration._setupDebugLoggingIfNeeded();
```
maios, line 8: is this ever called?

lib/src/log_configuration.dart
```diff
@@ -2,11 +2,57 @@ part of mapbox_maps_flutter;
 
 /// A class that allows to configure Mapbox SDKs logging per application.
 final class LogConfiguration {
+    // create an instance here to initialize debug logging once
+  static LogConfiguration _instance = LogConfiguration._();
```
maios, line 6: if I get this right, we want this to be lazy init, should it be with `late` keyword here?
also imho it's much easier with a static boolean flag indicating if the log writer has been registered, instead of doing a lazy loading here

maios (dismissed):

lib/src/mapbox_maps_options.dart
evil159: Ah, good point, it is never called indeed, I mistook it for `_MapboxOptions` constructor somehow, changed it to static variable init.

maios (dismissed):

maios (approved):

maios (approved):

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ final geometry = RenderedQueryGeometry(type: Type.SCREEN_COORDINATE, value jsonE
 final geometry = RenderedQueryGeometry.fromScreenCoordinate(screenCoordinate);
 ```
 * Bump Maps SDK to 11.7.0.
+* Print to console native Maps SDK logs in debug configuration.
```
pjleonard37, line 30: Perhaps add a note explaining how to turn this off here.

PR #700, opened 2024-09-23
Improve style image error reporting

evil159:

This PR adds error throwing on iOS(Android is not susceptible to this) in case data -> UIImage conversion fails. Additionally, I changed `hasStyleImage` to use the dedicated method(`imageExists()`) from the platform side on iOS(Android already uses that).

### Pull request checklist:
 - [ ] Add a changelog entry.


pjleonard37 (approved):
Nice changes 👍

PR #715, opened 2024-10-02
Drop proxy binary messenger

maios:
 Drop `ProxyBinaryMessenger`, instead, setup channel with a messageChannelSuffix which is now supported by Pigeon

### Pull request checklist:
 - [ ] Add a changelog entry.


ios/Classes/EventHandler.swift
```diff
@@ -45,12 +45,13 @@ final class MapboxEventHandler {
     private let channel: FlutterMethodChannel
     private var cancelables = Set<AnyCancelable>()
 
-    init(eventProvider: EventProvider, binaryMessenger: FlutterBinaryMessenger, eventTypes: [Int]) {
+    init(eventProvider: EventProvider, binaryMessenger: FlutterBinaryMessenger, eventTypes: [Int], channelSuffix: String = "") {
```
evil159, line 48: Channel suffix should be a required parameter here.

android/src/main/kotlin/com/mapbox/maps/mapbox_maps/snapshotter/SnapshotterInstanceManager.kt
```diff
@@ -34,9 +31,8 @@ class SnapshotterInstanceManager(
       options = options.toSnapshotOptions(context),
       overlayOptions = options.toSnapshotOverlayOptions()
     )
-    val proxyBinaryMessenger = ProxyBinaryMessenger(messenger, suffix)
     val styleManager: com.mapbox.maps.StyleManager = snapshotter.styleManager() // TODO: expose this on Android
-    val eventHandler = MapboxEventHandler(styleManager, proxyBinaryMessenger, eventTypes.map { it })
+    val eventHandler = MapboxEventHandler(styleManager, messenger, eventTypes.map { it }, "snapshot")
```
evil159, line 35: Here should be a non-static suffix

lib/src/offline/offline_manager.dart
evil159, line 1: I think this was an issue before as well - `EventChannel` created in `loadStylePack` does not have any suffixes - meaning it will broadcast to any listeners loading the same style uri, better attach some suffix to it as well.

lib/src/offline/tile_store.dart
evil159, line 1: Here in `loadTileRegion` we create `EventChannel` with an id of the region, I think in case of multiple tile store instances those ids could clash, safer to add a suffix there as well.

android/src/main/kotlin/com/mapbox/maps/mapbox_maps/EventHandler.kt
```diff
@@ -37,7 +37,8 @@ import java.util.Date
 class MapboxEventHandler(
   private val eventProvider: Observable,
   binaryMessenger: BinaryMessenger,
-  eventTypes: List<Long>
+  eventTypes: List<Long>,
+  channelSuffix: String = "",
```
evil159, line 41: This should be a required parameter.

lib/src/mapbox_maps_platform.dart
```diff
@@ -6,11 +6,18 @@ class _MapboxMapsPlatform {
   late final MethodChannel _channel = MethodChannel(
```
evil159, line 6: We have to add suffix to the name of this channel otherwise all kinds of events could end up broadcasting across all active map instances.

lib/src/snapshotter/snapshotter.dart
```diff
@@ -45,10 +45,10 @@ final class Snapshotter {
     this.onStyleDataLoadedListener,
     this.onStyleImageMissingListener,
   }) {
-    final messenger = ProxyBinaryMessenger(suffix: _suffix.toString());
-    _snapshotterMessenger = _SnapshotterMessenger(binaryMessenger: messenger);
-    style = StyleManager(binaryMessenger: messenger);
-    _mapEvents = _MapEvents(binaryMessenger: messenger);
+    _snapshotterMessenger =
+        _SnapshotterMessenger(messageChannelSuffix: _suffix.toString());
```
evil159, line 49: I noticed a pattern where binary messenger is specified explicitly, for consistency it should be specified here as well.

lib/src/map_events.dart
```diff
@@ -40,9 +40,13 @@ final class _MapEvents {
     return listenersMap.values.toList();
   }
 
-  _MapEvents({BinaryMessenger? binaryMessenger}) {
-    _channel = MethodChannel('com.mapbox.maps.flutter.map_events',
-        const StandardMethodCodec(), binaryMessenger);
+  _MapEvents({BinaryMessenger? binaryMessenger, String channelSuffix = ''}) {
```
evil159, line 43: Additionally binary messenger should be a required parameter as well, since in all other places binary messenger is specified explicitly, having it as a requirement will help with consistency.

evil159 (changes requested):
There some aspects to improve, mainly when it comes to multi-instance scenarios, but looking good overall.

lib/src/annotation/polyline_annotation_manager.dart
```diff
@@ -3,12 +3,14 @@ part of mapbox_maps_flutter;
...
-            _PolylineAnnotationMessenger(binaryMessenger: messenger),
-        super._();
+  PolylineAnnotationManager(
+      {required String id,
+      required BinaryMessenger messenger,
+      String? channelSuffix})
+      : super(id: id, messenger: messenger, channelSuffix: channelSuffix);
 
-  final _PolylineAnnotationMessenger _annotationMessenger;
+  late _PolylineAnnotationMessenger messenger = _PolylineAnnotationMessenger(
```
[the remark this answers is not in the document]
evil159: I setup linter to produce an error for cases like this(private types in public API), but it seems that those errors don't fail the build, I guess we'll have check it on CI in order to catch them.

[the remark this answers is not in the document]
evil159: Here's the PR https://github.com/mapbox/mapbox-maps-flutter/pull/735

lib/src/snapshotter/snapshotter.dart
maios: I don't think it is necessary, as stated by the doc, so we can leave it as optional null
> The [binaryMessenger] named argument is available for dependency injection.  If it is left null, the default BinaryMessenger will be used which routes to the host platform.

ios/Classes/MapboxMapFactory.swift
```diff
@@ -27,53 +27,35 @@ final class MapboxMapFactory: NSObject, FlutterPlatformViewFactory {
...
             withFrame: frame,
             mapInitOptions: mapInitOptions,
-            channelSuffix: channelSuffix,
-            arguments: args,
+            channelSuffix: args["channelSuffix"] as? Int ?? 0,
             registrar: registrar,
-            pluginVersion: pluginVersion,
-            eventTypes: eventTypes
+            pluginVersion: args["mapboxPluginVersion"] as? String ?? "",
+            eventTypes: args["eventTypes"] as? [Int] ?? []
```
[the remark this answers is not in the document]
maios: it's the same as before 😅 this one is only clean up the code a bit

pjleonard37 (approved):
Looks good from my side. I think ready to go once we get @evil159 's approval

lib/src/annotation/annotation_manager.dart
```diff
@@ -75,14 +75,12 @@ class AnnotationManager {
 
 /// The super class for all AnnotationManagers.
 class BaseAnnotationManager {
-  BaseAnnotationManager(
+  BaseAnnotationManager._(
       {required String id,
       required BinaryMessenger messenger,
-      String? channelSuffix})
+      required String channelSuffix})
```
evil159, line 81: nit: `channelSuffix` is not stored here anymore, might as well be removed from here. Subclasses might as well declare their own constructor parameter: `required super.channelSuffix` > `required String channelSuffix`)

lib/src/mapbox_maps_platform.dart
```diff
@@ -6,11 +6,18 @@ class _MapboxMapsPlatform {
...
       'plugins.flutter.io', const StandardMethodCodec(), binaryMessenger);
```
evil159, line 7: There should be suffix here, otherwise this method channel will broadcast all events to all active map instances.

evil159 (commented):
@maios thank you for addressing the feedback. I think we are almost there, I have left a couple of nits(in my opinion the binary messenger parameter should be removed), and I noticed there is one method channel in `_MapboxMapsPlatform` lacking the channel suffix.
Otherwise, it's looking very good!

P.S. This is an internal change, but its impact is broad. Should we note it in the changelog?

evil159 (approved):
Nicely done!

pjleonard37:
@maios  -- I think the `build-android` job is failing because of this unresolved reference to $messageChannel. 

https://github.com/mapbox/mapbox-maps-flutter/pull/715/files#diff-f72e4d76e69116de99f71b5e16fe8b73874d764869c822e44a81453325f7234dR100

Should this be `$channelSuffix`?

pjleonard37 (approved):

Issue #657, opened 2024-08-19
No implementation found for method subscribeToEvents


favazHF:
Error:

`Crash
MethodChannel._invokeMethod package:flutter/src/services/platform_channel.dart:332
MissingPluginException(No implementation found for method subscribeToEvents on channel com.mapbox.maps.flutter.map_events)`

Devices: Android and iOS

Mapbox version:  `mapbox_maps_flutter: ^2.1.0`

Wiwo99:
i have same problem!

PuntitOwO:
I was able to find a invocation in [map_events.dart:L58](https://github.com/mapbox/mapbox-maps-flutter/blob/3ce880a39153f0158f788f5e5bf4a6799cf856f7/lib/src/map_events.dart#L58C10-L58C27)
And it is supposed to handle it in [EventHandler.kt:L60](https://github.com/mapbox/mapbox-maps-flutter/blob/3ce880a39153f0158f788f5e5bf4a6799cf856f7/android/src/main/kotlin/com/mapbox/maps/mapbox_maps/EventHandler.kt#L60)
So, for some reason that condition is false 🤔

HighSoftWare96:
Same exception here (Android and library at 2.3.0-rc.1).. Anything on this? Which version is not affected?

mhammerc:
Hello, same problem here.

lukaskurz:
We are also experiencing this issue in our sentry logs from multiple devices, multiple OS

evil159:
Hi all, thank you for reporting this and sorry for the inconvenience, has anybody been able to reproduce it? A piece of sample code would be much appreciated 🙏

evil159:
I looked into this and so far my theory is that platform view might take a while get up and running, and during this time the channels are not hooked up, so if we will try to send event subscription update request then - we get an exception that nobody has answered the call.

I made a draft PR for this https://github.com/mapbox/mapbox-maps-flutter/pull/724 if somebody wants to try it out.

ciakov:
I have made some tests and the porblem looks that it is resolved.

favazHF:
@evil159 Do you think this solution will be released soon? the metrics of the error handlers are affected by this crash.

evil159:
@favazHF @ciakov Thank you for confirming the fix. It is scheduled for release in version 2.4.0, with the beta expected to be available later this week.

mauro-forest:
[...]

PR #724, opened 2024-10-09
Ignore event subsription updates while platform view is creating

evil159:

Platform view creation is asynchronous and in case widget state updates before the view is created we can get an exception trying to send a message to subscribe for events.

The fix is to ignore event subsription updates while platform view is creating.

Addresses https://github.com/mapbox/mapbox-maps-flutter/issues/657. 

### Pull request checklist:
 - [x] Add a changelog entry.


pjleonard37 (dismissed):
🎉

pjleonard37 (approved):
🌵

PR #738, opened 2024-10-17
Fix `GlyphsRasterizationMode` crashing `Snapshotter`

evil159:

Fixes crash in Snapshotter when GlyphsRasterizationMode is specified in MapSnapshotOptions. The crash was happening because Pigeon couldn't encode the enum, as the enum declaration was in another file. The fix was to add a reference to `GlyphsRasterizationMode` in the snapshotter template.

### Pull request checklist:
 - [x] Add a changelog entry.


pjleonard37 (approved):
Nice fix 👍

PR #748, opened 2024-10-24
Introduce Interactions API

pjleonard37:

This PR introduces the **experimental** Interactions API introduced on iOS and Android in 11.8.0-beta.1. Specifically it: 

- Introduces the experimental `MapboxMap.addInteractions` method, which allows you to add interactions to the map. 
  * Introduces `TapInteraction` and `LongTapInteraction`, which allow you to add tap and longTap interactions to the map.
  * Introduces `FeaturesetDescriptor` -- and convenience descriptors for `StandardBuildings`, `StandardPOIs`, and `StandardPlaceLabels` -- which allow you to describe the `Featureset`s you want `Interactions` to target.
- Introduces new methods for `queryRenderedFeatures`, `querySourceFeatures`, `setFeatureState`, `getFeatureState`, `removeFeatureState`, `resetFeatureState` and adds interfaces for iOS and Android 
- Adds all necessary extensions for conversions on iOS and Android 
- Adds an example showing how to query features in featuresets and set feature states. 
- Adds necessary Turf adapters   
- Adds tests 

https://github.com/user-attachments/assets/1bf2707a-2fe3-439a-99a6-a7ddd661c85f

Work for another PR: 
- Tests which use a mocked interaction 
- Expand on current example to better show use cases for API

- Platform alignment and enable Interactive Features 

### Pull request checklist:
 - [X] Add a changelog entry.


ios/Classes/Extensions.swift
```diff
@@ -418,6 +523,36 @@ extension MapboxMaps.MapOptions {
...
+            namespace: self.namespace)
+    }
+}
+
+extension MapboxMaps.FeaturesetDescriptor {
+    func toFLTFeaturesetDescriptor() -> FeaturesetDescriptor {
+        return FeaturesetDescriptor(
+            featuresetId: featuresetId,
+            importId: importId,
+            layerId: layerId)
```
pjleonard37, line 539: @persidskiy -- this conversion from `MapboxMaps.FeaturesetDescriptor` to Flutter's `FeaturesetDescriptor` fails because `featuresetId`, `importId`, and `layerId` are private (see [here](https://github.com/mapbox/mapbox-maps-ios-internal/blob/main/mapbox-maps-ios/Sources/MapboxMaps/Interactions/FeaturesetDescriptor.swift#L15)). Is it possible to make these public or expose them in a different way?

android/src/main/kotlin/com/mapbox/maps/mapbox_maps/Extentions.kt
```diff
@@ -500,8 +575,45 @@ fun com.mapbox.maps.QueriedFeature.toFLTQueriedFeature(): QueriedFeature {
...
+  val feature = Feature.fromGeometry(geometry, jsonObject)
+  properties.toMap()
+  val map: Map<String, Any?> = state.toMap()
+
+  return FeaturesetFeature(
+    id?.toFLTFeaturesetFeatureId(),
+    descriptor.toFeaturesetDescriptor().toFLTFeaturesetDescriptor(),
+    feature,
+    map
+  )
```
pjleonard37, line 616: @kiryldz -- For this conversion from `FeaturesetFeature<FeatureState>` to Flutter's `FeaturesetFeature` I'm having issues with converting the Feature itself. The originalFeature is internal (see [here](https://github.com/mapbox/mapbox-maps-android-internal/blob/cd3d94c4f7ec47bfbe8b1701be3fd2f43f60c4a0/mapbox-maps-android/sdk-base/src/main/java/com/mapbox/maps/interactions/FeaturesetFeature.kt#L40)) so I don't have access to that. I tried to create the Feature.fromGeometry, but that requires a JsonObject (see [here](https://github.com/mapbox/mapbox-java/blob/c75057c152dc121072363b5f0affe602cb1c64ea/services-geojson/src/main/java/com/mapbox/geojson/Feature.java#L130)) whereas FeaturesetFeature has a JSONObject (see [here](https://github.com/mapbox/mapbox-maps-android-internal/blob/cd3d94c4f7ec47bfbe8b1701be3fd2f43f60c4a0/mapbox-maps-android/sdk-base/src/main/java/com/mapbox/maps/interactions/FeaturesetFeature.kt#L53))*. I looked into conversions online, but they seemed very complex. What is the best way to approach this? Could originalFeature be made public? Could we use JsonObject instead for properties?  

`* It took me a few minutes to figure out those are different 😅

kiryldz: @pjleonard37 I am not sure I understand why do you actually need the `originalFeature`... The public API of `FeaturesetFeature` should be the same in Android and iOS and contain:
1. id: FeaturesetFeatureId? (nullable).
2. descriptor: TypedFeaturesetDescriptor (not nullable).
3. state: T (where T is generic, not nullable).
4. geometry: Geometry (not nullable, taken from GeoJSON lib in Android and from Turf in iOS)
5. properties: JSONObject (in Android `JSONObject` comes from Android SDK and is easier to use comparing to `JsonObject` coming from Google GSON library; not nullable).

So in Flutter I would like to see that you have the API surface I described above with some Flutter specifics (e.g. Android JSONObject should be mapped to some Flutter standard JSON, also I am sure in our Flutter SDK we already have a common equivalent for `Geometry` (representing point, line etc)). Then `originalFeature` does not seem to be needed.

pjleonard37: That's a good point. I was mirroring the internal implementation, but you're right that I do not need `originalFeature` at all. I've update the code in the to use geometry and properties instead (see [here](https://github.com/mapbox/mapbox-maps-flutter/pull/748/files#diff-f8388a280ca09b9e221cedabb19650472cc3ccb279fcc85dbf31f7e9ca5cd2f4R323)).

For geometry, I've represented it as `Map<String?, Object?>`, which is what we've used for other geometries in Flutter. We currently only have strongly typed `Point`, `Polygon`, and `LineString` (see [here](https://github.com/mapbox/mapbox-maps-flutter/blob/main/lib/src/turf_adapters.dart)). Though, we could expand that to cover all geometries, what do you think @evil159?

pjleonard37: > I think we can keep it as is and fix it separately, though changing APIs will be hard as these would mean breaking changes.
I agree with this approach. I started working on creating a `Geometry` type, but think it is beyond the scope of this PR. Since everything here is experimental, I think we should separate the `Geometry` work into a separate ticket and then we can make the change after.

ios/Classes/Extensions.swift
pjleonard37: Also, this is why the iOS build is failing here.

lib/src/mapbox_map.dart
```diff
@@ -426,11 +426,39 @@ class MapboxMap extends ChangeNotifier {
           _RenderedQueryGeometry(value: geometry.value, type: geometry.type),
           options);
 
+  /// Queries the map for rendered features using featureset descriptors.
+  Future<List<QueriedRenderedFeature?>> queryRenderedFeaturesForTargets(
```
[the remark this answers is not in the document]
pjleonard37: > I think Pigeon lifted this requirement recently.

Oh nice, it does not look like it needs to be optional, so I've updated the code in the latest commit.

lib/src/style/mapbox_styles.dart
```diff
@@ -36,6 +36,11 @@ class MapboxStyles {
   /// improve the style.
   static const String SATELLITE_STREETS =
       "mapbox://styles/mapbox/satellite-streets-v12";
+
+  /// NOT FOR PRODUCTION USE. An experimental version of the Mapbox Standard style.
+  /// This style is used for testing new features and changes to the Mapbox Standard style. The style may change or be removed at any time.
+  static const String STANDARD_EXPERIMENTAL =
```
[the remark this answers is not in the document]
pjleonard37: Updated in most recent commit.

ios/Classes/Extensions.swift
persidskiy: The idea is not to overexpose api that we don't need.

But I think getters can be exposed. Can you make a pr please?

example/lib/interactive_features_example.dart
```diff
@@ -0,0 +1,92 @@
...
+          bearing: 49.92,
+          zoom: 16.35,
+          pitch: 40),
+
+      /// DON'T USE Standard Experimental style in production, it will break over time.
+      /// Currently this feature is in preview.
+      styleUri: MapboxStyles.STANDARD_EXPERIMENTAL,
+      textureView: true,
+      onMapCreated: _onMapCreated,
+      onTapListener: _onTap,
```
persidskiy, line 88: can I add more than one tap listeners?

ios/Classes/Extensions.swift
pjleonard37: Makes sense. Yes, can make a PR 👍

example/lib/interactive_features_example.dart
persidskiy: @pjleonard37 Did you try to implement the `addInteraction` method? I think it makes sense to start with it, instead of QRF. 

With introduction of interactions, the Interaction is the main way to handle taps and we consider the qrf to be the low-level API. Users also will benefit from Interactions more, than from the qrf, because now they have a good cross-platform event handling system.

I think we should start with interactions and promote them to use Interactions in the examples.

example/lib/interactive_features_example.dart
```diff
@@ -0,0 +1,92 @@
...
+    var featureset =
+        FeaturesetDescriptor(featuresetId: "buildings", importId: "basemap");
+    var queriedFeatures = await mapboxMap?.queryRenderedFeaturesForFeatureset(
+        geometry: renderedQueryGeometry, featureset: featureset);
+    var featuresetFeatureId = queriedFeatures?.first.id;
+
+    // Remove that feature state
+    if (featuresetFeatureId != null) {
+      mapboxMap?.removeFeatureStateForFeaturesetFeatureDescriptor(
+          featureId: featuresetFeatureId, featureset: featureset);
```
persidskiy, line 68: let's have the featureset as first argument, to align with ios/android

persidskiy (commented):
some small bits of review

ios/Classes/Extensions.swift
pjleonard37: PR here: https://github.com/mapbox/mapbox-maps-ios-internal/pull/2344

example/lib/interactive_features_example.dart
```diff
@@ -0,0 +1,92 @@
...
+        geometry: renderedQueryGeometry, featureset: featureset);
+    var featuresetFeature = queriedFeatures?.first;
+
+    if (featuresetFeature != null) {
+      // Define the state to set for the feature, in this case highlighting
+      // Set that featurestate on that featuresetFeature
+      Map<String, Object?> state = {
+        "highlight": true,
+      };
+      mapboxMap?.setFeatureStateForFeaturesetFeature(featuresetFeature, state);
```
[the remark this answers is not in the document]
pjleonard37: Unfortunately, no. I agree that they are quite long, but I'm not sure how to shorten them without losing meaning. WDYT, @evil159 & @maios? 

QRF:

- queryRenderedFeatures(_RenderedQueryGeometry geometry, RenderedQueryOptions options)
  - This is the existing method  
- queryRenderedFeaturesForTargets(_RenderedQueryGeometry geometry, List<FeaturesetQueryTarget> targets)
- queryRenderedFeaturesForFeatureset( _RenderedQueryGeometry geometry, FeaturesetDescriptor featureset, String? filter)
- queryRenderedFeaturesInViewport(FeaturesetDescriptor featureset, String? filter)

QSF:
- querySourceFeatures(
      String sourceId, SourceQueryOptions options)
  - Existing method
- querySourceFeaturesForFeatureset(FeaturesetQueryTarget target)
  - Perhaps this could be querySourceFeaturesForTargets

Set/Get FeatureState: 
- set(get)FeatureState(
      String sourceId, String? sourceLayerId, String featureId, String state)
  - Existing method
- set(get)FeatureStateForFeaturesetDescriptor(
    FeaturesetDescriptor featureset, FeaturesetFeatureId featureId)
- set(get)FeatureStateForFeaturesetFeature(
    FeaturesetFeature feature, Map<String, Object?> state)

We could maybe drop "Featureset" from these ones 🤔

example/lib/interactive_features_example.dart
pjleonard37: > @pjleonard37 Did you try to implement the addInteraction method? I think it makes sense to start with it, instead of QRF.

I've experimented a bit locally, but wanted to keep the scope of this PR contained so I started with the low-level APIs. I do agree we should promote Interactions as the main entry point in our examples and docs.

persidskiy: Should we create a separate ticket for that, so we don't forget? In the https://mapbox.atlassian.net/browse/MAPSFLT-232 the `addInteraction` was in the scope.

android/src/main/kotlin/com/mapbox/maps/mapbox_maps/MapInterfaceController.kt
```diff
@@ -162,6 +176,72 @@ class MapInterfaceController(
...
+      callback(
+        Result.success(
+          it.map { feature -> feature.toFLTFeaturesetFeature() }.toMutableList()
+        )
+      )
+    }
+  }
+
+  @OptIn(MapboxExperimental::class)
+  override fun queryRenderedFeaturesInViewport(
```
kiryldz, line 226: I am not why this is a dedicated function. Moreover implementation should basically be calling `queryRenderedFeaturesForFeatureset` with `RenderedQueryGeometry(ScreenBox(ScreenCoordinate(0.0, 0.0), ScreenCoordinate(mapView.width.toDouble(), mapView.height.toDouble())))` instead of copying this impl

lib/src/mapbox_map.dart
```diff
@@ -426,11 +426,43 @@ class MapboxMap extends ChangeNotifier {
...
+          RenderedQueryGeometry geometry,
+          List<FeaturesetQueryTarget> targets) async =>
+      _mapInterface.queryRenderedFeaturesForTargets(
+          _RenderedQueryGeometry(value: geometry.value, type: geometry.type),
+          targets);
+
+  /// Queries the map for rendered features with one typed featureset.
+  @experimental
+  Future<List<FeaturesetFeature>> queryRenderedFeaturesForFeatureset(
```
persidskiy, line 440: The big part of iOS and Android SDK is the introduction of generated types specific for StandardStyle, and the generic-based system for this. The similar call in IOS leverages this system and returns you a feature type with corresponding type, e.g something like this:

```
let poiFeatures = await map.queryRenderedFeaturesForFeatureset(geom, FeaturesetDescriptor.standardPoi)
// poiFeatures here is a List<StandardPoiFeature>
```

If we can express this in Dart, I'd recommend doing this because it makes the api much more usable. WDYT?

example/lib/interactive_features_example.dart
```diff
@@ -0,0 +1,92 @@
...
+class InteractiveFeaturesState extends State<InteractiveFeaturesExample> {
+  InteractiveFeaturesState();
+  MapboxMap? mapboxMap;
+
+  _onMapCreated(MapboxMap mapboxMap) {
+    this.mapboxMap = mapboxMap;
+    mapboxMap.style;
+  }
+
+  _onTap(context) async {
```
kiryldz, line 26: For me it looks a bit strange that we're advertising the following approach of building highlight on a click. Honestly for me it looks more of an anti-approach comparing with Android / iOS 😄 The main reason for this is that in this example user defines their own order of click events, e.g.:

1. There's a POI on top of the building.
2. User clicks the POI.
3. User takes the code from this example and first implements `queryRenderedFeaturesForFeatureset` for buildings and then for POIs.
4. Then although actually POI was clicked - based on this code the building will be highlighted first.

I've just came from implementing interactions in Dash and it took me quite a while to re-write their logic with QRFs to interactions.

Bottom line: if we do not have time to add `addInteraction` - I'd at least modify this example and remove `onTap` at all to showcase highlighting the building for given coordinate. In the future Flutter versions we could prepare the proper example like:

```
_onTap(context) async {
 // map clicked
}

_onPoiTap(poi, context) async {
  // e.g. hide POI
}

_onBuildingTap(building, context) async {
  // highlight building
}
```

cc @persidskiy

android/src/main/kotlin/com/mapbox/maps/mapbox_maps/Extentions.kt
```diff
@@ -263,6 +270,74 @@ fun RenderedQueryOptions.toRenderedQueryOptions(): com.mapbox.maps.RenderedQuery
...
+        }
+        else -> throw (RuntimeException("Unsupported (key, value): ($key, $value)"))
+      }
+    }
+  }
+}
+
+@OptIn(MapboxExperimental::class)
+@SuppressLint("RestrictedApi")
+fun FeaturesetFeature.toFeaturesetFeature(): com.mapbox.maps.interactions.FeaturesetFeature<FeatureState> {
```
kiryldz, line 323: @pjleonard37 if I understand it right, this function is not needed. From what I see you use it internally only to map Flutter `FeaturesetFeature` to the Android one. However in Android (and most likely in iOS) we provide low-level API for `get / set / remove FeatureState` that take `TypedFeaturesetDescriptor` and `FeaturesetFeatureId` as parameters instead of `FeaturesetFeature`.

If you'll use those - you can remove `toFeaturesetFeature()` from what I see and also get rid of those nasty JSON conversions (which are actually not needed at all).

kiryldz: example: third overload [here](https://docs.mapbox.com/android/maps/api/11.8.0-rc.1/mapbox-maps-android/com.mapbox.maps/-mapbox-map/set-feature-state.html?query=fun%20%3CFS%20:%20FeatureState%3E%20setFeatureState(descriptor:%20TypedFeaturesetDescriptor%3CFS,%20*%3E,%20id:%20FeaturesetFeatureId,%20state:%20FS,%20callback:%20FeatureStateOperationCallback%20=%20FeatureStateOperationCallback%20{%20}):%20Cancelable) is the one that I would like to be used in Flutter

android/src/main/kotlin/com/mapbox/maps/mapbox_maps/MapInterfaceController.kt
pjleonard37: I added this as iOS has it (see [here](https://github.com/mapbox/mapbox-maps-ios-internal/blob/bca5e7c559e25d6465a81905f5fad38d1e613bd5/mapbox-maps-ios/Sources/MapboxMaps/Foundation/MapboxMap.swift#L1148)) and do think it is a nice convenience for developers. If we'd like to keep it, I can switch the implementation. 

CC: @persidskiy

example/lib/interactive_features_example.dart
pjleonard37: Yes -- that ticket should cover `Map.addInteraction`, `TapInteraction`, `LongPressInteraction`, and generated derivative features for Standard Style (like `StandardBuilding`, `StandardPoiLabel`), which I'll create. As for the work itself, as we will not be able to merge this PR into main until a new iOS maps with [this](https://github.com/mapbox/mapbox-maps-ios-internal/pull/2344) lands anyways, it will be in a new PR into this branch.

lib/src/mapbox_map.dart
pjleonard37: > If we can express this in Dart

That's what I'm trying to figure out in the ongoing work: how to express this in Dart in a way that is (1) handles all necessary conversions across the platforms, (2) can be generated by our pigeons, and (3) creates a consistent public API with the iOS and Android APIs.

example/lib/interactive_features_example.dart
pjleonard37: That's a fair point -- the example should more clearly demonstrate the logic we'd like the developer to think in. I can rewrite this today.

android/src/main/kotlin/com/mapbox/maps/mapbox_maps/Extentions.kt
pjleonard37: Interesting, I'll need to think about this a bit 🤔 

>  get rid of those nasty JSON conversions
This would be great 👍

example/lib/interactive_features_example.dart
persidskiy: Yeah, this PR implements the low-level API - QRF, but the most useful now is the `addInteraction()`.

Ideally, we would expose it here too, and used it in the example

android/src/main/kotlin/com/mapbox/maps/mapbox_maps/MapInterfaceController.kt
kiryldz: I think we should remove it from iOS. Implementation seems very straight-forward (use rect 0;0 -> width; height for another similar QRF). At the same time this method does not take into account camera padding which makes it incorrect for Navigation Camera scenario (as they always use padding).

persidskiy: This is just a simplified version of qrf, very useful when you want just get something that is on screed but filtered by a filter.

In GLJS, the `geometry` parameter is also optional, and does the same. In iOS it's the different method only because we have a generic Geometry. See https://docs.mapbox.com/mapbox-gl-js/api/map/#map#queryrenderedfeatures

In flutter, ideally we can the geometry optional

kiryldz: @persidskiy thanks for more info. WDYT if I'll expose this method on Android similar to JS (for QRF with TypedFeaturesetDescriptor only as it's experimental)? 

image

persidskiy: Sure, that would be perfect 👍

kiryldz: Android implementation: https://github.com/mapbox/mapbox-maps-android-internal/pull/2831

example/lib/interactive_features_example.dart
```diff
@@ -0,0 +1,77 @@
...
+
+class InteractiveFeaturesState extends State<InteractiveFeaturesExample> {
+  InteractiveFeaturesState();
+  MapboxMap? mapboxMap;
+
+  _onMapCreated(MapboxMap mapboxMap) {
+    this.mapboxMap = mapboxMap;
+    mapboxMap.style;
+
+    /// Define interactions for 3D Buildings
```
pjleonard37, line 25: Note that these building interactions will not work with the STANDARD style. Use STANDARD_EXPERIMENTAL instead to test.

ios/mapbox_maps_flutter/Sources/mapbox_maps_flutter/Classes/MapboxMapController.swift
```diff
@@ -106,6 +106,38 @@ final class MapboxMapController: NSObject, FlutterPlatformView {
...
+                  let interactionType = InteractionType(rawValue: interactionTypeRaw),
+                  let stopPropagation = arguments["stopPropagation"] as? Bool,
+                  let id = arguments["id"] as? Int else {
+                return
+            }
+            let filter = arguments["filter"] as? String
+            let filterExpression = try? filter.flatMap { try $0.toExp() }
+            let radius = arguments["radius"] as? CGFloat
+
+            switch interactionType {
```
pjleonard37, line 124: I feel that this could be simplified somehow, but I can't think of exactly how. Recommendations are most welcome.

lib/src/mapbox_map.dart
```diff
@@ -505,6 +552,50 @@ class MapboxMap extends ChangeNotifier {
...
+          FeaturesetDescriptor featureset) =>
+      _mapInterface.resetFeatureStatesForFeatureset(featureset);
+
+  // References for all interactions added to the map.
+  @experimental
+  var interactionsList = _InteractionsList(interactions: {});
+
+  /// Add an interaction
+  @experimental
+  void addInteraction(Interaction interaction, OnInteraction action) {
```
pjleonard37, line 587: @evil159  -- I'd appreciate a particularly thorough review of this section (setting up the listener and message channel. I model it broadly after our approach to gestures, but would welcome any suggested improvements. What do you think of this system of using both `_InteractionListener` and `_InteractionsList`?

lib/src/mapbox_maps_platform.dart
```diff
@@ -161,6 +161,23 @@ class _MapboxMapsPlatform {
     }
   }
 
+  Future<dynamic> addInteractionsListeners(
+      Interaction interaction, int interactionID) async {
+    try {
+      return _channel
+          .invokeMethod('interactions#add_interaction', <String, dynamic>{
```
pjleonard37, line 168: @evil159  - any better approach here you can think of?

lib/src/style/mapbox_styles.dart
```diff
@@ -36,6 +36,12 @@ class MapboxStyles {
   /// improve the style.
   static const String SATELLITE_STREETS =
       "mapbox://styles/mapbox/satellite-streets-v12";
+
+  /// NOT FOR PRODUCTION USE. An experimental version of the Mapbox Standard style.
+  /// This style is used for testing new features and changes to the Mapbox Standard style. The style may change or be removed at any time.
+  @experimental
```
pjleonard37, line 42: This can be removed, I've just left it in for now for convenience when reviewing.

lib/src/style/interactive_features/standard_poi.dart
```diff
@@ -0,0 +1,83 @@
+part of mapbox_maps_flutter;
+
+// A feature that is a point of interest in the Standard style.
+class StandardPoiFeature extends FeaturesetFeature {
```
pjleonard37, line 4: @persidskiy and @evil159  - I've introduced these Standard-specific types here which also have corresponding `FeatureState`s which can be used like this: 

```
  StandardBuildingState featureState = StandardBuildingState(highlight: true);

  mapboxMap.addInteraction(tapInteraction, (_, FeaturesetFeature feature) {
    mapboxMap.setFeatureStateForFeaturesetFeature(feature, featureState);
    var buildingFeature = StandardBuildingsFeature(
        feature.geometry, feature.properties, feature.state,
        id: feature.id);
    print("Building feature id: ${buildingFeature.id}");
  });
```

I'd like to push the concept farther and return them instead of the generic `FeaturesetFeature`, but I'm running into some limits of what we can do with our Dart and Pigeon setup. I can't think of how we'd pass a typed version of `FeaturesetFeature` with `OnInteraction`. Any ideas? 

```
typedef void OnInteraction(
    MapContentGestureContext context, FeaturesetFeature feature);
```

lib/src/pigeons/map_interfaces.dart
```diff
@@ -1289,6 +1307,257 @@ class QueriedFeature {
...
+  });
+
+  /// The featureset descriptor that specifies the featureset to be included in the interaction.
+  FeaturesetDescriptor featuresetDescriptor;
+
+  /// The type of interaction, either tap or longTap
+  InteractionType interactionType;
+
+  /// Whether to stop the propagation of the interaction to the map. Defaults to true.
+  bool stopPropagation;
```
pjleonard37, line 1390: Not that I've included this here rather than in the callback

android/src/main/kotlin/com/mapbox/maps/mapbox_maps/MapInterfaceController.kt
```diff
@@ -253,6 +284,34 @@ class MapInterfaceController(
...
+    ) { callback(Result.success(Unit)) }
+  }
+
+  @OptIn(MapboxExperimental::class, MapboxDelicateApi::class)
+  override fun setFeatureStateForFeaturesetFeature(
+    feature: FeaturesetFeature,
+    state: Map<String, Any?>,
+    callback: (Result<Unit>) -> Unit
+  ) {
+    val id = feature.id?.toFeaturesetFeatureId() ?: return
```
evil159, line 307: We can't return without invoking the callback - dart will hang invoking async execution context until reply is delivered.

android/src/main/kotlin/com/mapbox/maps/mapbox_maps/MapInterfaceController.kt
```diff
@@ -270,6 +329,45 @@ class MapInterfaceController(
     }
   }
 
+  @OptIn(MapboxExperimental::class, MapboxDelicateApi::class)
+  override fun getFeatureStateForFeaturesetDescriptor(
+    featureset: FeaturesetDescriptor,
+    featureId: FeaturesetFeatureId,
+    callback: (Result<Map<String, Any?>>) -> Unit
+  ) {
+    featureset.toTypedFeaturesetDescriptor()?.let { featuresetDescriptor ->
```
evil159, line 338: The callback will not be invoked if conversion fails.

android/src/main/kotlin/com/mapbox/maps/mapbox_maps/MapInterfaceController.kt
```diff
@@ -270,6 +329,45 @@ class MapInterfaceController(
...
+    }
+  }
+
+  @OptIn(MapboxExperimental::class, MapboxDelicateApi::class)
+  override fun getFeatureStateForFeaturesetFeature(
+    feature: FeaturesetFeature,
+    callback: (Result<Map<String, Any?>>) -> Unit
+  ) {
+    val featuresetDescriptor = feature.featureset.toTypedFeaturesetDescriptor() ?: return
+    val id = feature.id?.toFeaturesetFeatureId() ?: return
```
evil159, line 358: The callback has to be invoked when exiting early.

android/src/main/kotlin/com/mapbox/maps/mapbox_maps/MapInterfaceController.kt
```diff
@@ -286,6 +384,62 @@ class MapInterfaceController(
...
+    }
+  }
+
+  @OptIn(MapboxExperimental::class, MapboxDelicateApi::class)
+  override fun removeFeatureStateForFeaturesetFeature(
+    feature: FeaturesetFeature,
+    stateKey: String?,
+    callback: (Result<Unit>) -> Unit
+  ) {
+    val id = feature.id?.toFeaturesetFeatureId() ?: return
```
evil159, line 413: Early return without invoking the callback.

ios/mapbox_maps_flutter/Sources/mapbox_maps_flutter/Classes/MapInterfaceController.swift
```diff
@@ -284,6 +286,40 @@ final class MapInterfaceController: _MapInterface {
         }
     }
 
+    func setFeatureStateForFeaturesetDescriptor(featureset: FeaturesetDescriptor, featureId: FeaturesetFeatureId, state: [String: Any?], completion: @escaping (Result<Void, any Error>) -> Void) {
+        guard let state = JSONObject.init(turfRawValue: state) else {
+            return
```
evil159, line 291: Early return without invoking the completion.

ios/mapbox_maps_flutter/Sources/mapbox_maps_flutter/Classes/MapInterfaceController.swift
```diff
@@ -284,6 +286,40 @@ final class MapInterfaceController: _MapInterface {
...
+                )))
+            } else {
+                completion(.success(()))
+            }
+        }
+    }
+
+    func setFeatureStateForFeaturesetFeature(feature: FeaturesetFeature, state: [String: Any?], completion: @escaping (Result<Void, Error>) -> Void) {
+        guard let state = JSONObject.init(turfRawValue: state) else {
+            return
```
evil159, line 308: Early return without invoking the completion.

lib/src/mapbox_map.dart
evil159: Given the limitations - I think this is perfectly fine way to do it(it is also probably the only way to do it :)

evil159: I think the only thing that is missing here is the ability to remove interaction.

lib/src/style/interactive_features/standard_poi.dart
evil159: Seems that the type system allows this:
```dart
interface class MyInterface {}
class MyImplementation implements MyInterface {
  String? bar = "bar";
}
class MyImplementation2 implements MyInterface {
  String? baz = "baz";
}

class MyInteraction<T extends MyInterface> {
  final T impl;
  MyInteraction(this.impl);
}


void foo<T extends MyInterface>(MyInteraction<T> i, void Function(T) callback) {
  callback(i.impl);
}

void main() {
  final i = MyInteraction(MyImplementation());
  foo(i, (MyImplementation i) {
    print(i.bar);
  });

  final i2 = MyInteraction(MyImplementation2());
  foo(i2, (MyImplementation2 i) {
    print(i.baz);
  });
}
```

lib/src/style/interactive_features/interactive_features.dart
```diff
@@ -0,0 +1,44 @@
...
+      : super(
+            featuresetDescriptor: featuresetDescriptor,
+            interactionType: InteractionType.TAP,
+            filter: filter,
+            radius: radius,
+            stopPropagation: stopPropagation);
+}
+
+// A long tap interaction
+class LongTapInteraction extends Interaction {
```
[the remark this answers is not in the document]
pjleonard37: I debated about naming here and decided to align with our gesture naming for Flutter, which is what the individual platforms do. Ideally, I think we'd align across all platforms, but we'd have to pick one of the two here. I thought it was better to be internally consistent, but happy to change if you think otherwise. 

Flutter: 
- [Gestures](https://github.com/mapbox/mapbox-maps-flutter-internal/blob/e0d6be9faea8b7b4e0f9124cccdae41dad62918b/codegen/pigeons/gesture_listeners.dart#L48): onTap; onLongTap; onScroll

iOS: 
- [Gestures](https://github.com/mapbox/mapbox-maps-ios-internal/blob/6b89816365f6eb4a72833eb1edc68137800d28c7/mapbox-maps-ios/Sources/MapboxMaps/Gestures/GestureManager.swift#L61): singleTap; longPress; pan
- [Interactions](https://github.com/mapbox/mapbox-maps-ios-internal/blob/6b89816365f6eb4a72833eb1edc68137800d28c7/mapbox-maps-ios/Sources/MapboxMaps/Interactions/Interactions.swift#L38): Tap; LongPress; Drag

Android: 
- [Gestures](https://github.com/mapbox/mapbox-maps-android-internal/blob/08145594c6386a4a449b214df9911f38493720fc/mapbox-maps-android/sdk-base/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPlugin.kt#L44): onMapClick; onMapLongClick; onMove
- [Interactions](https://github.com/mapbox/mapbox-maps-android-internal/tree/main/mapbox-maps-android/sdk-base/src/main/java/com/mapbox/maps): Click; LongClick; Drag

android/src/main/kotlin/com/mapbox/maps/mapbox_maps/MapInterfaceController.kt
pjleonard37: Good catch -- I've updated all of these

lib/src/mapbox_map.dart
```diff
@@ -505,6 +552,50 @@ class MapboxMap extends ChangeNotifier {
...
+          {required FeaturesetDescriptor featureset,
+          required FeaturesetFeatureId featureId,
+          String? stateKey}) =>
+      _mapInterface.removeFeatureStateForFeaturesetDescriptor(
+          featureset, featureId, stateKey);
+
+  /// Removes entries from a specified Feature.
+  /// Remove a specified property or all property from a feature's state object, depending on the value of `stateKey`.
+  @experimental
+  Future<void> removeFeatureStateForFeaturesetFeature(
```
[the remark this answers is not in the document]
evil159: The difference is that it is opt-in behavior on platform side, here, the error is propagated unconditionally. We could pipe those errors to the returned future, this way it would replicate the platform setup - those interested in errors could subscribe to them with `Future.onError(...)`, and I guess normal use case would be to ignore them by default.

lib/src/style/interactive_features/interactive_features.dart
```diff
@@ -0,0 +1,44 @@
+part of mapbox_maps_flutter;
```
persidskiy, line 1: @pjleonard37 Is this the manually written or generated code? I think we need to take a similar approach to Android/iOS SDK and generate this.

lib/src/style/interactive_features/interactive_features.dart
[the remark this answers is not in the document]
persidskiy: @pjleonard37 On android/iOS we decided to follow the platform patterns instead of aligning so it will be more natural on each platform.

According to this APIhttps://api.flutter.dev/flutter/widgets/GestureDetector/onLongPress.html, shouldn't we ineed name thie `LongPressInteraction`?

lib/src/style/interactive_features/interactive_features.dart
```diff
@@ -0,0 +1,44 @@
...
+            featuresetDescriptor: featuresetDescriptor,
+            interactionType: InteractionType.LONG_TAP,
+            filter: filter,
+            radius: radius,
+            stopPropagation: stopPropagation);
+}
+
+extension Featureset on FeaturesetDescriptor {
+  // A Featureset of buildings in the Standard style
+  static FeaturesetDescriptor standardBuildings({String importId = "basemap"}) {
```
persidskiy, line 29: Can we take a similar approach to Android/iOS and generate this?

example/assets/featuresetsStyle.json
persidskiy, line 1: IS this used for tests? I wonder why it is in the examples folder?

example/lib/interactive_features_example.dart
```diff
@@ -0,0 +1,77 @@
...
+
+    // Define a tap interaction targeting the Buildings featureset in the Standard style
+    var tapInteraction = TapInteraction(Featureset.standardBuildings());
```
persidskiy, line 28: On iOS/Android the Interaction itself consumes the callback upon construction, but in Flutter the callback is supplied into `addInteraction` method. Is it some limitation?

lib/src/style/interactive_features/standard_poi.dart
```diff
@@ -0,0 +1,83 @@
+part of mapbox_maps_flutter;
```
persidskiy, line 1: In Android/iOS we started with manual implementation of those classes, but quickly realized that there's no way to align on it, if we don't generate them from the style.

Can we generate them as well in Flutter?

persidskiy (commented):
@pjleonard37 The PR looks good. I left some comments here and there touching different design choices

lib/src/style/interactive_features/interactive_features.dart
```diff
@@ -0,0 +1,44 @@
...
+      {String? filter, double? radius, bool stopPropagation = true})
+      : super(
+            featuresetDescriptor: featuresetDescriptor,
+            interactionType: InteractionType.LONG_TAP,
+            filter: filter,
+            radius: radius,
+            stopPropagation: stopPropagation);
+}
+
+extension Featureset on FeaturesetDescriptor {
```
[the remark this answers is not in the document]
pjleonard37: 👍  That would be nice. We could also short the name of the static descriptors so for implementation if would look like: 

LongTapInteraction(Standard.Buildings() ...

lib/src/style/interactive_features/interactive_features.dart
pjleonard37: I'd like to, but didn't want to expand the scope of this PR further. I can create a ticket.

example/assets/featuresetsStyle.json
pjleonard37: Originally I has used this as part of an example, but I've now moved it into the test folder instead.

pjleonard37: The integration tests are run from the examples so there assets are in the same folder (see [here](https://github.com/mapbox/mapbox-maps-flutter/tree/ff141671fecce93524fbfe28b2072a79a37b306a/example)).

lib/src/style/interactive_features/interactive_features.dart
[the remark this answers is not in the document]
pjleonard37: I'm happy to use `LongPress`, but think we should then update our gesture name in the next Flutter major release so we're aligned internally. What do you think @evil159 & @maios?

lib/src/mapbox_map.dart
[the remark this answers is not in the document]
pjleonard37: Ahhh, I see -- ok, I'll need to think about the best way here.

lib/src/pigeons/map_interfaces.dart
```diff
@@ -1289,6 +1307,257 @@ class QueriedFeature {
...
+      geometry: (result[2] as Map<Object?, Object?>?)!.cast<String?, Object?>(),
+      properties:
+          (result[3] as Map<Object?, Object?>?)!.cast<String, Object?>(),
+      state: (result[4] as Map<Object?, Object?>?)!.cast<String, Object?>(),
+    );
+  }
+}
+
+/// Defines the parameters for querying features from a Featureset with an optional filter and id.
+class FeaturesetQueryTarget {
```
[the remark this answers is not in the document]
pjleonard37: Removed in most recent commit.

lib/src/pigeons/map_interfaces.dart
```diff
@@ -1205,10 +1215,16 @@ class QueriedRenderedFeature {
   /// If the feature is only rendered in one layer, a single Id will be provided.
   List<String?> layers;
 
+  /// An array of feature query targets that correspond to this queried feature.
+  ///
+  /// - Note: Returned query targets will omit the original `filter` data.
+  List<FeaturesetQueryTarget>? queryTargets;
```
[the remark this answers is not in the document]
pjleonard37: Removed in most recent commit

lib/src/mapbox_map.dart
```diff
@@ -505,6 +552,50 @@ class MapboxMap extends ChangeNotifier {
...
+      _mapInterface.resetFeatureStatesForFeatureset(featureset);
+
+  // References for all interactions added to the map.
+  @experimental
+  var interactionsList = _InteractionsList(interactions: {});
+
+  /// Add an interaction
+  @experimental
+  void addInteraction(Interaction interaction, OnInteraction action) {
+    interactionsList.interactions[action.hashCode] = _InteractionListener(
```
persidskiy, line 588: From the hashCode documentation:
```
/// Objects that are not equal are allowed to have the same hash code.
```
So we cannot use hashcode as a key in the dictionary.

lib/src/mapbox_map.dart
```diff
@@ -505,6 +552,50 @@ class MapboxMap extends ChangeNotifier {
...
+      onInteractionListener: action,
+      interactionID: action.hashCode,
+    );
+
+    InteractionsListener.setUp(interactionsList,
```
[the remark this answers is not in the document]
pjleonard37: We need to pass the list with all interactions, but there is no way to know when the final interaction has been added (unlike gestures where we have a limited set so we can just set up once). This approach overwrites the channel with the new list of all interactions each time one is added so that we can be sure messages can be sent through all of them. Though correct me if I'm thinking about channels incorrectly, @evil159.

This `.setup()` is generated by pigeon, so we do not have much flexibility in how we implement it.

example/lib/interactive_features_example.dart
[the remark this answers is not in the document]
pjleonard37: Good catch. I looked into this more and the issue is not actually with interactions, but feature state (including the version in our production code). The feature is being properly set as `highlight: true`, but the map is not redrawn on Android (it is on iOS). This means that visually the map looks the same. You can test this by triggering the feature state change, waiting for it to apply, and then panning the map. Immediately the building will highlight when panning. We should definitely address this. I'll cut a ticket. 

[Screen_recording_20250116_160833.webm](https://github.com/user-attachments/assets/a6f40228-123d-40ec-9c3a-ee394964b249)

lib/src/mapbox_map.dart
```diff
@@ -505,6 +552,50 @@ class MapboxMap extends ChangeNotifier {
...
+
+  // References for all interactions added to the map.
+  @experimental
+  var interactionsList = _InteractionsList(interactions: {});
+
+  /// Add an interaction
+  @experimental
+  void addInteraction(Interaction interaction, OnInteraction action) {
+    interactionsList.interactions[action.hashCode] = _InteractionListener(
+      onInteractionListener: action,
```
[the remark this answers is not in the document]
pjleonard37: The behavior in Flutter matches the platform behavior. But that behavior is not as you describe. It varies based on the featureset type. Standard Place Labels & Standard POIs work as you describe, and the last added interaction executes first. Standard Buildings is the opposite with the first added triggering first. You can test this in the iOS SDK example: 

```
LongPressInteraction(.standardBuildings) { _, _ in
    print("Buildings 1")
    return false
}
LongPressInteraction(.standardPoi) { _, _  in
    print("POIs 1")
    return false
}
LongPressInteraction(.standardBuildings) { _, _ in
    print("Buildings 2")
    return false
}
LongPressInteraction(.standardPoi) { _, _  in
    print("POIs 2")
    return false
}
LongPressInteraction(.standardBuildings) { _, _  in
    print("Buildings 3")
    return false
}
LongPressInteraction(.standardPoi) { _, _  in
    print("POIs 3")
    return false
}
```

If you long press (or tap) on a POI over a building this will print: 

> POIs 3
> POIs 2
> POIs 1
> Buildings 1
> Buildings 2
> Buildings 3

I replicated this on Android as well. I'm not sure why this is, and we should ticket for investigation.

lib/src/mapbox_maps_platform.dart
pjleonard37: I think that's a good improvement -- I'll ticket it

example/lib/interactive_features_example.dart
```diff
@@ -0,0 +1,77 @@
...
+
+    // Define a state to highlight the building when it is interacted with
+    StandardBuildingState featureState = StandardBuildingState(highlight: true);
+
+    // Add the tap interaction to the map, set the action to occur when a building is tapped (highlight it)
+    mapboxMap.addInteraction(tapInteraction, (_, FeaturesetFeature feature) {
```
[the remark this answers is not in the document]
pjleonard37: With the latest commits I found a way for the Interaction to have to callback, rather than having it on `addInteraction` 👍

lib/src/pigeons/map_interfaces.dart
```diff
@@ -168,6 +168,15 @@ enum ViewAnnotationAnchor {
   CENTER,
 }
 
+/// The type of interaction, either tap/click or longTap/longClick
+enum InteractionType {
```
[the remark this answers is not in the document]
pjleonard37: Updated to internal 👍

lib/src/style/interactive_features/standard_poi.dart
pjleonard37: Yes: I agree on generation and will ticket

lib/src/mapbox_map.dart
pjleonard37: I've changed this to the order of addition

lib/src/mapbox_maps_platform.dart
pjleonard37: I tried this out in the latest commit with the creation of `_InteractionPigeon`. I'm curious to hear your thoughts on this. I think it cleans up the platform code just a bit.

lib/src/mapbox_map.dart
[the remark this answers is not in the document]
evil159: Looks very good now!

example/lib/interactive_features_example.dart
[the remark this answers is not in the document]
evil159: @pjleonard37, is there a ticket for this?

lib/src/mapbox_maps_platform.dart
evil159: Hmmm... I don't see the reason to wrap interaction into `_InteractionPigeon` anymore(not sure why I didn't suggest it in the first place, sorry), we can just pass `_Interaction` itself, right?

lib/src/mapbox_map.dart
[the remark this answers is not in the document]
persidskiy: Interresting, looks like a bug for buildings. 
In the POI labels the output is correct.

example/lib/interactive_features_example.dart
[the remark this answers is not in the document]
pjleonard37: Yes: MAPSFLT-289

lib/src/style/interactive_features/interactive_features.dart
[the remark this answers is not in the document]
pjleonard37: 👍 -- sounds good. I'll keep Tap and LongTap

lib/src/mapbox_maps_platform.dart
pjleonard37: We can't pass `_Interaction` because it has types which are not supported by [MessageCodec](https://api.flutter.dev/flutter/services/StandardMessageCodec-class.html). Specifically: `FeaturesetDescriptor` and `_InteractionType`.

lib/src/pigeons/map_interfaces.dart
persidskiy: Today in offline discussion with @evil159 I realized that we cannot implement it at all due to async communication dart<->platform. So this configuration totally makes sense, resolving this convo 👍

lib/src/mapbox_map.dart
```diff
@@ -675,3 +768,53 @@ class _GestureListener extends GestureListener {
...
+    } else {
+      typedFeature = feature as T;
+    }
+
+    onInteractionListener.call(context, typedFeature);
+  }
+}
+
+/// Listen to all interactions on the map, determine which interaction to call
+class _InteractionsList<T extends FeaturesetFeature>
```
persidskiy, line 807: nit: _InteractionsMap or _InteractionsContainer would be a better name, since it is implemented as Map, not as List internally.

persidskiy (dismissed):
LGTM 🔥

only two essential notes:
- we need a way to remove interacitons
- it would be nice to remove the need to specify Feature type explicitly upon interaction creation.

lib/src/style/interactive_features/standard_poi.dart
pjleonard37: Updated here: https://github.com/mapbox/mapbox-maps-flutter/pull/748#discussion_r1932742744

lib/src/mapbox_maps_platform.dart
evil159: I see, they've changed the way objects are encoded/decoded lately, you can make it work passing the `_Interaction` type to a pigeon-generated method, but yeah, it won't work out of box when using custom channels.

evil159 (approved):
I have to admit I didn't go through the whole PR after recent changes, instead I've reviewed the parts that were changed since and they look good, thank you @pjleonard37 for the effort!

example/lib/interactive_features_example.dart
```diff
@@ -0,0 +1,84 @@
...
+    mapboxMap.style;
+
+    /// Define interactions for 3D Buildings
+
+    // Define a state to highlight the building when it is interacted with
+    StandardBuildingState featureState = StandardBuildingState(highlight: true);
+
+    // Define a tap interaction targeting the Buildings featureset in the Standard style
+    // Set the action to occur when a building is tapped (highlight it)
+    var tapInteraction = TapInteraction<StandardBuildingsFeature>(
```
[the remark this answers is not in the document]
persidskiy: Awesome 👍

PR #837, opened 2025-01-21
Update README.md

shinriyo:
it is not nullable
and it is Point object

fix old one

I want good README

### Pull request checklist:
 - [ ] Add a changelog entry.


pjleonard37 (approved):
Thanks for these edits!

PR #839, opened 2025-01-21
Update versions for v2.6.0-beta.1 release 

pjleonard37:

This PR updates necessary versions for the v2.6.0-beta.1 release. In particular, note that the iOS minimum is now 14.0. Also note that an experimental import was necessary to use experimental features.

### Pull request checklist:
 - [ ] Add a changelog entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 # main
 
+### 2.6.0-beta.1
+
+> [!IMPORTANT]
+> The iOS minimum deployment target is now iOS 14.0.
+
 * Add support for Swift Package Manager.
+* Update Maps SDK to 11.10.0-beta.1
 
 ### 2.5.0
 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Mapbox Maps SDK Flutter SDK
 
-The Mapbox Maps SDK Flutter SDK is an officially developed solution from Mapbox that enables use of our latest Maps SDK product (v11.9.0). The SDK allows developers to embed highly-customized maps using a Flutter widget on Android and iOS.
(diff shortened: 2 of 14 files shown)


evil159 (dismissed):

evil159 (approved):

mauriciolcs:
According to the Mapbox [v11.9.0](https://github.com/mapbox/mapbox-maps-ios/releases/tag/v11.9.0) change log, the minimum iOS version is 14.0 since Mapbox v11.9.1. So the latest version of the mapbox-maps-flutter needs a fix to increase the minimum version.

pjleonard37:
@mauriciolcs -- 

> The first two lines [here](https://github.com/mapbox/mapbox-maps-ios/releases/tag/v11.9.0) warn that the next versions will have iOS version 14 as the minimum version.

I can see how that wording could be confusing. I've updated the comment to clarify. 

> Therefore, I believe that version 11.10 requires version 14 as the minimum version, so much so that when I update mapbox-maps-flutter to version 2.5.1, I get an error due to the minimum iOS version, as shown in the screenshot.

Maps iOS 11.10 does require version 14 as the minimum version. However, Maps Flutter 2.5.1 uses Maps iOS 11.9.1, which supports versions 12.0 and 13.0.

mauriciolcs:
@pjleonard37 ah, sorry, I ended up writing the wrong version in my previous message.

I meant that it seems that version 11.9.1 also needs to have iOS 14 as a minimum, since I'm having trouble running the plugin on version 2.5.1 using iOS 13 as a minimum.

pjleonard37:
Hi @mauriciolcs - 

Thanks for clarifying. I just tried on my end and had no issues. Could you try again? If you are still running into issues could you then open a new issue [here](https://github.com/mapbox/mapbox-maps-flutter/issues) with a simple project that replicates so we can investigate?

PR #852, opened 2025-01-30
Add removeInteraction, map targeting to Interactions API

pjleonard37:

1.) Adds the ability to remove interaction. To remove an interaction, a developer can pass an optional interaction ID when adding the interaction, like this: 

```
mapboxMap.addInteraction(tapInteractionPOI, interactionID: "tap_interaction_poi");
```

The interaction can then be removed like this: 

```
mapboxMap.removeInteraction("tap_interaction_poi");
```

This PR additionally updates the example to showcase removing the POI interaction. 

2.) Add the ability to add an interaction targeting the map itself. As no feature is returned this required adding some additional callback typedefs. An example implementation was added to the main InteractiveFeatures example. 

```
var longTapInteraction = LongTapInteraction.onMap((context) {
  log("Long tap at: ${context.touchPosition.x}, ${context.touchPosition.y}");
}, stopPropagation: false);
```

3.) Align ordering of feature, context with the iOS and Android platforms. 

Note: iOS will not build until the next release of iOS which includes updated access for `.dispatch()`. 

Alignment with platform APIs and better developer ergonomics. 

### Pull request checklist:
 - [X] Add a changelog entry.


lib/src/mapbox_map.dart
```diff
@@ -580,14 +580,16 @@ class MapboxMap extends ChangeNotifier {
...
+  /// Add an interaction to the map
+  /// An identifier can be provided, which you can use to remove
+  /// the interaction with `.removeInteraction(interactionID)`
   @experimental
   void addInteraction<T extends TypedFeaturesetFeature<FeaturesetDescriptor>>(
-      TypedInteraction<T> interaction) {
-    var id = _interactionsMap.interactions.length;
+      TypedInteraction<T> interaction,
+      {String? interactionID}) {
+    var id = interactionID ?? DateTime.now().microsecondsSinceEpoch.toString();
```
pjleonard37, line 592: @evil159 - Is there a better way to assign a unique string?

android/src/main/kotlin/com/mapbox/maps/mapbox_maps/MapboxMapController.kt
```diff
@@ -242,50 +238,11 @@ class MapboxMapController(
...
-          }
-        }
+        interactionsController.addInteraction(messenger, channelSuffix, call)
+        result.success(null)
+      }
+      "interactions#remove_interaction" -> {
+        interactionsController.removeInteraction(call)
         result.success(null)
       }
       "platform#releaseMethodChannels" -> {
```
pjleonard37, line 248: I don't need to do anything with `dispose()` here, right?

lib/src/mapbox_map.dart
```diff
@@ -580,14 +584,16 @@ class MapboxMap extends ChangeNotifier {
...
```
maios, line 596: is it gonna mutate?
```suggestion
    final id = interactionID ?? DateTime.now().microsecondsSinceEpoch.toString();
```

lib/src/mapbox_map.dart
maios: Can we use UniqueKey here?https://api.flutter.dev/flutter/foundation/UniqueKey-class.html

lib/src/mapbox_map.dart
```diff
@@ -777,33 +791,37 @@ class _InteractionListener<T extends FeaturesetFeature>
...
-      typedFeature =
-          TypedFeaturesetFeature<StandardPOIs>.fromFeaturesetFeature(feature)
-              as T;
-    } else if (featuresetID == "place-labels") {
-      typedFeature =
-          TypedFeaturesetFeature<StandardPlaceLabels>.fromFeaturesetFeature(
-              feature) as T;
+  void onInteraction(FeaturesetFeature? feature,
+      MapContentGestureContext context, String interactionID) {
+    var featuresetID = feature?.featureset.featuresetId;
```
maios, line 801:
```suggestion
    final featuresetID = feature?.featureset.featuresetId;
```

pjleonard37: 👍  -- updated in latest commit

lib/src/mapbox_map.dart
pjleonard37: 👍  -- updated in latest commit

lib/src/mapbox_map.dart
pjleonard37: 👍  -- updated in latest commit

.fastlane/Fastfile
```diff
@@ -30,7 +30,7 @@ platform :ios do
       scheme: 'Runner',                   # XCTest scheme
       xcconfig: "example/ios/Flutter/Release.xcconfig",
       derived_data_path: "example/build/output",
-      configuration: "Release",
+      configuration: "Debug",
```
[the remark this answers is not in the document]
pjleonard37: This has now been reverted. It was an attempt to be able to use `.dispatch()` without exposing it, but we'll need to wait for [this PR](https://github.com/mapbox/mapbox-maps-ios-internal/pull/2447) to land and be released instead.

android/src/main/kotlin/com/mapbox/maps/mapbox_maps/MapboxMapController.kt
evil159: Unless you need to perform some cleanup upon map disposal

evil159 (approved):

android/src/main/kotlin/com/mapbox/maps/mapbox_maps/MapboxMapController.kt
pjleonard37: I couldn't think of anything -- no 👍

evil159 (approved):

PR #865, opened 2025-02-13
Update Standard Style Interactions Examples, Fix Bug

pjleonard37:

This PR updates the Standard Style Interactions Examples to align with iOS and Android. 

1.) standard_style_import_example was added to show how to import fragments and add interactions (like in this [example](https://docs.mapbox.com/android/maps/examples/android-view/add-interaction-to-custom-featuresets/)). However, we are missing the `.addStyleImport()` API, so this will need to be added when that is exposed. 

2.) interactive_features_example was renamed to standard_style_interactions_example and modified to allow changing the building color and other import configurations (like [here](https://docs.mapbox.com/android/maps/examples/android-view/add-interaction-to-featuresets/)). 

https://github.com/user-attachments/assets/cd116fcb-954d-49f7-a328-65c173d160ab

Additionally, this PR fixes a bug with `.setStyleImportConfigProperties()` where values were not being properly converted which led to the property not actually being updated. 

Better explain how to use these APIs and align with iOS/ Android 

### Pull request checklist:
 - [X] Add a changelog entry.


example/lib/main.dart
```diff
@@ -42,7 +43,8 @@ final List<Example> _allPages = <Example>[
   SpinningGlobeExample(),
   FullMapExample(),
   StyleExample(),
-  InteractiveFeaturesExample(),
+  StandardStyleImportExample(),
+  StandardStyleInteractionsExample(),
```
evil159, line 47: Move these two above full map and style examples?

evil159 (dismissed):

evil159 (approved):

PR #871, opened 2025-02-19
Update link for location puck examples

sohaibnomanahmed:
The link for examples related to the puck examples seems outdated and sends you to a page that does not exist. I found that the link was updated in this page https://docs.mapbox.com/flutter/maps/guides/user-location/. Since many users might just go into the README page, it should probably be updated here too


pjleonard37:
Thanks for the correction. Closing as the change has been made here: https://github.com/mapbox/mapbox-maps-flutter/blob/main/README.md

PR #876, opened 2025-02-26
Update versions for 2.6.1, include layer updates

pjleonard37:

This pull requests updates all required versions for the v2.6.1 release. Additionally, it includes required changes to `LocationIndicatorLayer`. Specifically, the `top-image`, `bearing-image`, and `shadow-image` properties on `LocationIndicatorLayer` are now paint properties instead of layout properties. 

### Pull request checklist:
 - [x] Add a changelog entry.


example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -199,6 +199,7 @@
 				97C146EC1CF9000F007C117D /* Resources */,
 				9705A1C41CF9048500538489 /* Embed Frameworks */,
 				3B06AD1E1E4923F5004D2608 /* Thin Binary */,
+				8C7C6621B70C9CF94E219643 /* [CP] Embed Pods Frameworks */,
```
maios, line 202: this should not be included since we are moving away from Cocoapods to use SPM, can you turn on SPM support and build the examples app again?

maios (approved):

example/ios/Runner.xcodeproj/project.pbxproj
pjleonard37: Good catch -- updated in latest commit :)

PR #891, opened 2025-03-12
Prepare release v2.7.0 rc.1

maios:


evil159:
@maios Seems that dependency version in `Package.swift` is not updated to RC

pjleonard37:
I've updated Package.swift. Could you take another look @evil159 ?

PR #914, opened 2025-04-15
Implement annotation drag events listener

maios:

Add listener callbacks for annotation's drag begin/changed/ended

[Screen_recording_20250415_105755.webm](https://github.com/user-attachments/assets/4fd1c559-e6f1-472f-9fb9-7ea99141435f)

### Pull request checklist:
 - [ ] Add a changelog entry.


example/integration_test/style/layer/model_layer_test.dart
```diff
@@ -27,7 +27,6 @@ void main() {
       modelColor: Colors.red.value,
       modelColorMixIntensity: 1.0,
       modelCutoffFadeRange: 1.0,
-      modelElevationReference: ModelElevationReference.SEA,
```
pjleonard37, line 30: Was this removed intentionally?

example/lib/edit_polygon_example.dart
```diff
@@ -0,0 +1,109 @@
...
+          }
+        ]
+      };
+      await mapboxMap.style.addSource(
+          GeoJsonSource(id: sourceId, data: jsonEncode(geoJsonData)));
+
+      await mapboxMap.style.addLayer(FillLayer(
+        id: "city-fill-layer",
+        sourceId: "city-fill-source",
+        fillColor: Colors.amberAccent.value,
```
pjleonard37, line 97: Nit: change this to a color that contrasts more with the land

pjleonard37 (changes requested):
I added a few comments from an initial review. Overall the example works well! Can you add integration tests and a changelog? 

I'll take a deeper look then.

example/integration_test/all_test.dart
```diff
@@ -51,6 +51,16 @@ import 'viewport_test.dart' as viewport_test;
...
+  // annotation tests
+  circle_annotation_manager_test.main();
+  circle_annotation_test.main();
+  point_annotation_manager_test.main();
+  point_annotation_test.main();
+  polygon_annotation_manager_test.main();
+  polygon_annotation_test.main();
+  polyline_annotation_manager_test.main();
+  polyline_annotation_test.main();
+
```
maios, line 63: I moved these tests up here since I'm fixing the suffix 0 for mocking event channel

ios/mapbox_maps_flutter/Sources/mapbox_maps_flutter/Classes/Annotations/AnnotationController.swift
```diff
@@ -131,6 +168,12 @@ class AnnotationController: ControllerDelegate {
         onCircleAnnotationClickListener = OnCircleAnnotationClickListener(binaryMessenger: binaryMessenger.messenger, messageChannelSuffix: binaryMessenger.suffix)
         onPolygonAnnotationClickListener = OnPolygonAnnotationClickListener(binaryMessenger: binaryMessenger.messenger, messageChannelSuffix: binaryMessenger.suffix)
         onPolylineAnnotationClickListener = OnPolylineAnnotationClickListener(binaryMessenger: binaryMessenger.messenger, messageChannelSuffix: binaryMessenger.suffix)
+
+        AnnotationDragEventsStreamHandler.register(
```
evil159, line 172: This will have to be setup per annotation manager instance I believe

android/src/main/kotlin/com/mapbox/maps/mapbox_maps/annotation/AnnotationController.kt
```diff
@@ -112,6 +174,21 @@ class AnnotationController(private val mapView: MapView) :
...
+      override fun onListen(p0: Any?, sink: PigeonEventSink<AnnotationInteractionContext>) {
+        interactionEventsSink = sink
+      }
+
+      override fun onCancel(p0: Any?) {
+        interactionEventsSink = null
+      }
+    }
+
+    AnnotationDragEventsStreamHandler.register(
```
evil159, line 188: Same point about setting this up once per annotation manager instance here

evil159 (commented):
Example looks 🔥 Nice tests as well! I left some points about multi-instance/unsubscribing support.

lib/src/annotation/circle_annotation_manager.dart
```diff
@@ -22,6 +22,55 @@ class CircleAnnotationManager extends BaseAnnotationManager {
...
+  ///   onEnd: (context) {
+  ///     print("Drag ended at: ${context.annotation.geometry}");
+  ///   },
+  /// );
+  /// ```
+  void dragEvents({
+    Function(CircleAnnotationInteractionContext)? onBegin,
+    Function(CircleAnnotationInteractionContext)? onChanged,
+    Function(CircleAnnotationInteractionContext)? onEnd,
+  }) {
```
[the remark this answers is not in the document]
maios: I updated to return an instance of `AnyCancelable` to cancel this subscription, note that the stream will not be canceled, only the subscription is canceled

example/lib/edit_polygon_example.dart
```diff
@@ -58,51 +68,41 @@ class EditPolygonExampleState extends State<EditPolygonExample> {
...
+
+    await mapboxMap.style.addLayer(FillLayer(
+      id: "city-fill-layer",
+      sourceId: _sourceId,
+      fillColor: Colors.pink.value,
+      fillOpacity: 0.3,
+    ));
   }
 
   Future<void> _addOrUpdateFillLayer() async {
```
evil159, line 100: This method needs to be renamed now(and the if condition should not be needed anymore since the source is setup in onStyleLoaded) or maybe it could be inlined since it's a one-liner now?

lib/src/cancelable.dart
```diff
@@ -5,15 +5,15 @@ part of '../mapbox_maps_flutter.dart';
 /// This can be used to manage operations that may need to be
 /// canceled before completion, such as asynchronous tasks or
 /// long-running processes.
-class AnyCancelable {
+class Cancelable {
   final void Function() cancel;
```
evil159, line 9: Using this is a bit weird since the autocomplete suggest just plain property `myCancellable.cancel`, so I have to          append `()` myself.
I think a public `cancel()` function would be more familiar and ergonomic. In this case autocomplete would suggests exactly what user intended.

maios: ah you're right, I've updated to have public cancel() function :)

evil159 (approved):
🔥

pjleonard37 (approved):
🔥🔥🔥 Looks great now!

PR #944, opened 2025-05-15
Public interface camera api

maios:

- Moved `ScreenCoordinate`, `MbxEdgeInsets`, `CameraOptions` and `CameraState` to `mapbox_maps_flutter_interface`
- Allow camera to be set upon Map initization, and add API to get/set camera.
- Provide web implementations
### Pull request checklist:
 - [ ] Add a changelog entry.


mapbox_maps_flutter_mobile/example/integration_test/animation_test.dart
```diff
@@ -3,6 +3,9 @@ import 'dart:io';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:integration_test/integration_test.dart';
 import 'package:mapbox_maps_flutter/mapbox_maps_flutter.dart';
+import 'package:mapbox_maps_flutter_interface/mapbox_maps_flutter_interface.dart'
```
evil159, line 6: Not sure if this is the way to go - the examples need to import the public package not the internal one.

mapbox_maps_flutter_mobile/lib/src/mapbox_map.dart
```diff
@@ -344,10 +344,16 @@ class MapboxMap extends ChangeNotifier {
   /// The map will retain its current values for any details not passed via the camera options argument.
   /// It is not guaranteed that the provided `camera options` will be set, the map may apply constraints resulting in a
   /// different `camera state`.
-  Future<void> setCamera(CameraOptions cameraOptions) =>
-      _cameraManager.setCamera(cameraOptions);
+  @override
+  Future<void> setCamera(CameraOptions cameraOptions) {
+    if (cameraOptions.center != null) {
+      cameraOptions.center = Point._from(cameraOptions.center!);
```
evil159, line 350: Why does this have to be re-assigned?

mapbox_maps_flutter_mobile/lib/src/mapbox_map.dart
```diff
@@ -458,6 +464,7 @@ class MapboxMap extends ChangeNotifier {
           options);
 
   /// Queries the map for rendered features with one typed featureset.
+  @experimental
```
evil159, line 467: Is tagging them experimental intentional?

maios: nope, I have no idea why it got here, will remove

mapbox_maps_flutter_mobile/lib/src/mapbox_map.dart
maios: initially I left the part where we extend turf's Point in mobile package, that's why I had to cast type here so the Codec would understand, otherwise it would throw an error Invalid type. But now I moved those extensions to interface package and removed this cast

PR #947, opened 2025-05-16
Fix ci v3

maios:


mapbox_maps_flutter_mobile/LICENSE
```diff
@@ -75,31 +75,6 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ```
```
pjleonard37, line 77: Is removing this intentional?

pjleonard37 (approved):

mapbox_maps_flutter_mobile/LICENSE
maios: yes, turf has been moved to mapbox_maps_flutter_interface, however, before publishing these new packages we need to discuss which one should be included in the license file.

Issue #791, opened 2024-11-29
Annotation click listeners - handling multiple annotations tapped at once


chris-wickens:
When zoomed out far enough for multiple point annotations to overlap, tapping the points results in separate callbacks to the OnPointAnnotationClickListener. I'd like to provide the option for the user to select which point they intended to tap, but the SDK doesn't provide the information of which points were tapped in one go. The only way I can think of doing this is to introduce an artificial delay to wait for each callback to be invoked, but I don't like the idea as the SDK already has a noticeable delay between the tap and the callback and I don't want to increase it further. Is there a better approach?

maios:
Hi @chris-wickens, thank you for reporting this. I created an internal issue to track the progress https://mapbox.atlassian.net/browse/MAPSFLT-192 
We plan to update MapboxMaps Flutter to use platforms's v11 implementation of map content gestures, where only the top annotation will report the tap.

alastairrmcneill:
@chris-wickens  did you ever come up with a solution for this? 

@maios any update on this ticket?

yunpaik:
@maios can you provide any update on this ticket please?

maios:
Hi, we have introduced [Interactive Map Elements](https://docs.mapbox.com/flutter/maps/guides/user-interaction/interactions/#add-an-interaction-to-the-map-itself) since MapboxMaps Flutter v2.6.0.

So you can add a TapInteraction to the layer hosting annotations, set `stopPropagation` to true to avoid triggering tap event of the annotation below, this makes sure that only the top annotation will get tap.
```
final tapInteraction = TapInteraction(
  FeaturesetDescriptor(layerId: value.id),
  (feature, context) {
    print("Tapped on annotation with id: ${feature.id}");
  },
  stopPropagation: true, // don't propagate to annotation below
);

mapboxMap.addInteraction(
  tapInteraction,
  interactionID: "tap_interaction_circle_annotation",
);
```

And if you want to get a report on multiple annotations being tapped and present the users which one they want to tap, you can do so with a TapInteraction on map, and use QueriedRenderedFeature API to find out the list of annotations at touch point

```
final tapInteraction = TapInteraction.onMap((context) async {
  final features = await mapboxMap.queryRenderedFeaturesForFeatureset(
    featureset: FeaturesetDescriptor(layerId: value.id),
    geometry: RenderedQueryGeometry.fromScreenCoordinate(
      context.touchPosition,
    ),
    filter: null,
  );
[...]

PR #977, opened 2025-06-24
Annotation tap and long press events

maios:

This PR adds `tapEvents` and `longPressEvents` to Annotations API, allowing users to provide a callback to be invoked and an annotation is tapped or long pressed.
Both API will return a Cancelable token, which can be used to stop listening to interaction events.

With `tapEvents`, we are now aligning with platform implementation that use IME underlying, instead of using old API in iOS and observe interaction events via delegation, and since the tap interaction is now not propagated to the annotation below the topmost one, you will get exactly one callback when you tap on overlapping annotations as described in #791 

### Pull request checklist:
 - [x] Add a changelog entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,9 @@
     },
   );
   ```
+
 > [!NOTE]
+> As part of this change, `AnnotationOnClickListener` is now deprecated.
 > Tap events will now not propagate to annotations below the topmost one. If you tap on overlapping annotations, only the top annotation's tap event will be triggered.
 
 ### 2.9.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,7 @@
 ### main
 
 * Introduce new experimental properties: `FillLayer.fillConstructBridgeGuardRail`, `FillLayer.fillBridgeGuardRailColor`, `FillLayer.fillTunnelStructureColor`, `CircleLayer.circleElevationReference`. 
-* Introduce `tapEvents` API to the Annotation Managers to handle tap event callbacks for annotations:
-  * `onTap`: Called when an annotation is tapped.
-
+* Introduce `tapEvents` and `longPressEvents` API to the Annotation Managers to handle tap and long press event callbacks for annotations:
   Example usage:
   ```dart
   manager.tapEvents(
@@ -11,9 +9,14 @@
       print("Tapped annotation: ${annotation.id}");
     },
   );
+  manager.longPressEvents(
+    onLongPress: (annotation) {
+      print("Long press annotation: ${annotation.id}");
+    },
+  );
   ```
-  > [!NOTE]
-  > Tap events will now not propagate to annotations below the topmost one. If you tap on overlapping annotations, only the top annotation's tap event will be triggered.
+> [!NOTE]
+> Tap events will now not propagate to annotations below the topmost one. If you tap on overlapping annotations, only the top annotation's tap event will be triggered.
 
 ### 2.9.0
 
--- a/example/lib/circle_annotations_example.dart
+++ b/example/lib/circle_annotations_example.dart
@@ -15,20 +15,6 @@ class CircleAnnotationExample extends StatefulWidget implements Example {
   State<StatefulWidget> createState() => CircleAnnotationExampleState();
 }
 
-class AnnotationClickListener extends OnCircleAnnotationClickListener {
-  AnnotationClickListener({
-    required this.onAnnotationClick,
-  });
-
-  final void Function(CircleAnnotation annotation) onAnnotationClick;
-
-  @override
-  void onCircleAnnotationClick(CircleAnnotation annotation) {
-    print("onAnnotationClick, id: ${annotation.id}");
-    onAnnotationClick(annotation);
-  }
-}
-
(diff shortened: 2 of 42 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,19 @@
...
+  Example usage:
+  ```dart
+  manager.tapEvents(
+    onTap: (annotation) {
+      print("Tapped annotation: ${annotation.id}");
+    },
+  );
+  ```
+  > [!NOTE]
+  > Tap events will now not propagate to annotations below the topmost one. If you tap on overlapping annotations, only the top annotation's tap event will be triggered.
```
evil159, line 16: Maybe it will be worth adding this note to the method documentation as well?

example/lib/circle_annotations_example.dart
```diff
@@ -55,11 +55,10 @@ class CircleAnnotationExampleState extends State<CircleAnnotationExample> {
         ));
       }
       circleAnnotationManager?.createMulti(options);
-      circleAnnotationManager?.addOnCircleAnnotationClickListener(
-        AnnotationClickListener(
-          onAnnotationClick: (annotation) => circleAnnotation = annotation,
-        ),
-      );
+      circleAnnotationManager?.tapEvents(onTap: (annotation) {
```
pjleonard37, line 58: I think we can now remove the AnnotationClickListener class from this example (and the other examples). It would be good to do this so devs are not confused.

lib/src/annotation/circle_annotation_manager.dart
```diff
@@ -16,10 +16,33 @@ class CircleAnnotationManager extends BaseAnnotationManager {
...
+  @Deprecated('Use [tapEvents] instead.')
   void addOnCircleAnnotationClickListener(
       OnCircleAnnotationClickListener listener) {
-    OnCircleAnnotationClickListener.setUp(listener,
-        binaryMessenger: _messenger, messageChannelSuffix: _channelSuffix);
+    OnCircleAnnotationClickListener._withCancelable(
+        tapEvents(onTap: listener.onCircleAnnotationClick), _channelSuffix);
+  }
+
+  /// Registers tap event callbacks for the annotations managed by this instance.
```
pjleonard37, line 26:
```suggestion
  /// Registers tap event callbacks for the annotations managed by this manager.
```

lib/src/annotation/point_annotation_manager.dart
```diff
@@ -469,11 +492,11 @@ class PointAnnotationManager extends BaseAnnotationManager {
   Future<double?> getIconHaloWidth() =>
       _annotationMessenger.getIconHaloWidth(id);
 
-  /// Controls the transition progress between the image variants of icon-image. Zero means the first variant is used, one is the second, and in between they are blended together. Both images should be the same size and have the same type (either raster or vector). Default value: 0. Value range: [0, 1]
+  /// Controls the transition progress between the image variants of icon-image. Zero means the first variant is used, one is the second, and in between they are blended together. . Both images should be the same size and have the same type (either raster or vector). Default value: 0. Value range: [0, 1]
```
pjleonard37, line 495: This is a typo upstream, but we should remove it here anyways. 

```suggestion
  /// Controls the transition progress between the image variants of icon-image. Zero means the first variant is used, one is the second, and in between they are blended together. Both images should be the same size and have the same type (either raster or vector). Default value: 0. Value range: [0, 1]
```

lib/src/annotation/point_annotation_manager.dart
```diff
@@ -469,11 +492,11 @@ class PointAnnotationManager extends BaseAnnotationManager {
...
   Future<void> setIconImageCrossFade(double iconImageCrossFade) =>
       _annotationMessenger.setIconImageCrossFade(id, iconImageCrossFade);
 
-  /// Controls the transition progress between the image variants of icon-image. Zero means the first variant is used, one is the second, and in between they are blended together. Both images should be the same size and have the same type (either raster or vector). Default value: 0. Value range: [0, 1]
+  /// Controls the transition progress between the image variants of icon-image. Zero means the first variant is used, one is the second, and in between they are blended together. . Both images should be the same size and have the same type (either raster or vector). Default value: 0. Value range: [0, 1]
```
pjleonard37, line 499:
```suggestion
  /// Controls the transition progress between the image variants of icon-image. Zero means the first variant is used, one is the second, and in between they are blended together. Both images should be the same size and have the same type (either raster or vector). Default value: 0. Value range: [0, 1]
```

CHANGELOG.md
```diff
@@ -1,6 +1,22 @@
 ### main
 
 * Introduce new experimental properties: `FillLayer.fillConstructBridgeGuardRail`, `FillLayer.fillBridgeGuardRailColor`, `FillLayer.fillTunnelStructureColor`, `CircleLayer.circleElevationReference`. 
+* Introduce `tapEvents` and `longPressEvents` API to the Annotation Managers to handle tap and long press event callbacks for annotations:
```
pjleonard37, line 4: Can we mention the deprecation here as well?

pjleonard37 (changes requested):
Overall looks really good! A few small edits and then I think it'll be ready to merge 🚀

pjleonard37 (approved):
🚀

PR #1081, opened 2025-12-10
Add method to delete multiple annotations

evil159:

This PR adds `deleteMulti()` to all annotation managers to allow deleting multiple annotations at once.

https://mapbox.atlassian.net/browse/MAPSFLT-390

### Pull request checklist:
 - [x] Add a changelog entry.

### DIFF
--- a/example/integration_test/annotations/circle_annotation_test.dart
+++ b/example/integration_test/annotations/circle_annotation_test.dart
@@ -71,5 +71,46 @@ void main() {
 
     await manager.deleteAll();
   });
+
+  testWidgets('deleteMulti CircleAnnotation', (WidgetTester tester) async {
+    final mapFuture = app.main();
+    await tester.pumpAndSettle();
+    final mapboxMap = await mapFuture;
+    final manager = await mapboxMap.annotations.createCircleAnnotationManager();
+    var geometry = Point(coordinates: Position(1.0, 2.0));
+
+    var circleAnnotationOptions = CircleAnnotationOptions(
+      geometry: geometry,
+    );
+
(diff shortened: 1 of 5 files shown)


example/lib/polyline_annotations_example.dart
```diff
@@ -30,15 +31,20 @@ class PolylineAnnotationExampleState extends State<PolylineAnnotationExample> {
       polylineAnnotationManager = value;
       createOneAnnotation();
       final positions = <List<Position>>[];
-      for (int i = 0; i < 99; i++) {
+      for (int i = 0; i < 500; i++) {
```
pjleonard37: 500 might be too many 😅 

Image

example/lib/polygon_annotations_example.dart
```diff
@@ -34,12 +35,15 @@ class PolygonAnnotationExampleState extends State<PolygonAnnotationExample> {
       polygonAnnotationManager = value;
       createOneAnnotation();
       var options = <PolygonAnnotationOptions>[];
-      for (var i = 0; i < 2; i++) {
+      for (var i = 0; i < 500; i++) {
```
pjleonard37: Similar to polyline, reduce the number so we can see a bit of the actual map 

Image

example/lib/polyline_annotations_example.dart
evil159: Sorry, checked only with circle annotations example and 500 seemed reasonable there, put the numbers back

pjleonard37 (approved):
Changes look good 👍 . We should generate tests as well

PR #1101, opened 2026-02-10
feat: add HTTP interceptor support

### LINKED ISSUE #1062: Support setting custom HTTP headers before Map initialization
We can set custom HTTP headers via setCustomHeaders(...), but we can only do that inside the onMapCreated callback (after obtaining the MapboxMap instance).
However, this timing is too late: the initial network requests (style / glyph / sprite / tile) are already fired before setCustomHeaders done, so those first requests never include our custom headers.

We’d like to define custom headers before any Maps API object is constructed — similar to how we set the access token with MapboxOptions.setAccessToken(...).

Rationale

Some servers require custom auth headers for all map traffic

Enterprise / B2B applications often need to identify the app / org via headers

Prevents the “first request goes out without headers” problem

Thanks!

dballance:

Adds HTTP interceptor support, in Dart, to match up to the native SDK functionality.

The interceptors are added before a map is instantiated, so that all requests (including fonts, images, etc) can be modified / logged / observed when the map is created. All map instances will share the same interceptor, much like the existing SDK functionality.

I'm opening this first to get feedback on the overall approach - will add tests if needed. Seems most of the native code is only tested indirectly, so would potentially just be an integration test of some sort.

Allow per request header modifications, logging, and tracing spans on map requests. Would potentially close #1062 and other stories.

### Pull request checklist:
 - [x] Add a changelog entry.


android/src/main/kotlin/com/mapbox/maps/mapbox_maps/MapboxMapsPlugin.kt
```diff
@@ -54,6 +57,48 @@ class MapboxMapsPlugin : FlutterPlugin, ActivityAware {
...
+          }
+        }
+        else -> {
+          result.notImplemented()
+        }
+      }
+    }
   }
 
   override fun onDetachedFromEngine(binding: FlutterPlugin.FlutterPluginBinding) {
```
pjleonard37: `httpInterceptorChannel` is set up in `onAttachedToEngine` but never torn down. So on hot restart the old method call handler
keeps running. You should add:

```kotlin
httpInterceptorChannel?.setMethodCallHandler(null)
httpInterceptorChannel = null
CustomHttpServiceInterceptor.getInstance().setFlutterChannel(null)
```
 here to `onDetachedFromEngine`.

example/integration_test/http_interceptor_test.dart
```diff
@@ -0,0 +1,371 @@
+import 'dart:async';
+
+import 'package:flutter_test/flutter_test.dart';
+import 'package:integration_test/integration_test.dart';
+import 'package:mapbox_maps_flutter/mapbox_maps_flutter.dart';
+import 'empty_map_widget.dart' as app;
+
+void main() {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
```
pjleonard37: Let's also clear the cache here: 

```dart
setUp(() async {
    await MapboxMapsOptions.clearData();
});
```

dballance:
@pjleonard37  - I updated this shortly after the last comment - is something else missing here?

dballance:
Hoping we can get this into a 3.0.0 release!

Issue #602, opened 2024-06-29
Compass not moving in the new verison of mapbox-maps flutter 


alladyaditi:
I cannot move the compass and see it , i was able to see it with this code
void _updateOrnamentSettings() {
    _mapboxMap.compass.updateSettings(CompassSettings(
      position: compassPosition,
      enabled: showOrnaments,
      fadeWhenFacingNorth: false,
      marginBottom: 10,
      marginLeft: 10,
      marginTop: 10,
      marginRight: 10,
    ));
    logger.i("Compass settings updated: position - $compassPosition, enabled - $showOrnaments");
  }

maios:
@alladyaditi Could you elaborate what happened, maybe best with a video showing the problems?

O-Hannonen:
Closing this as there is no further clarification and a related PR was merged, thanks for reporting! 

In case the PR did not fix your issue, feel free to re-open with more details

PR #1125, opened 2026-05-27
fix(ios): apply fadeWhenFacingNorth independently of enabled

charles-plante:
On iOS, CompassSettings.fadeWhenFacingNorth was ignored unless enabled was also set. CompassController.updateSettings only assigned compass.visibility inside `if let visible = settings.enabled { ... }`, so a call that set only fadeWhenFacingNorth never reached the native API and the compass stayed .adaptive, fading out at north. Android applies each field independently, so the platforms behaved differently.

Apply enabled and fadeWhenFacingNorth independently, reading the current visibility for whichever field the caller didn't provide, mirroring the Android CompassMappings implementation.

Related to #602.


charles-plante:
image

The license is signed

evil159 (approved):
@charles-plante Thank you for your contribution!

evil159:
Imported in 993430ae583844a41684253bf0de104797d4e071

PR #1143, opened 2026-08-28
feat(ios): render the map into a flutter texture instead of a platform view

### LINKED ISSUE #1142: iOS: the map cannot be captured or read by flutter, because it is a platform view
### what happens

on ios the map is a `UiKitView`. uikit composites it, not flutter, so the map is not in flutter's scene. anything that needs to read the map back therefore reads nothing:

* `RepaintBoundary.toImage` and `toImageSync` return a hole where the map is. screenshot a page with a map on it and you get everything except the map.
* a custom `ImageFilter.shader` used as a backdrop samples an empty texture, so a shader driven surface over the map renders as whatever it does with no input. in our case that was solid black.
* anything else that composites against what is beneath it has nothing beneath it.

plain `BackdropFilter` is fine, the engine special cases it, so this is not "you cannot blur the map". it is narrower and harder: you cannot get the map's pixels into flutter.

android does not have this problem. `MapWidget` there can render into a `TextureView`, which puts the map inside the flutter scene where everything can see it. there is no equivalent on ios.

### why it is not just a flag we forgot to port

android's `textureView` switches the render surface inside a platform view that already composites in the flutter scene. on ios platform views are always uikit composited (flutter/flutter#43902), so the fix is not a flag, it is giving the map a real texture path.

### how to reproduce

1. put a `MapWidget` in a page
2. wrap the page in a `RepaintBoundary`
3. call `toImage` on it and draw the result

the map area comes back blank. everything drawn by flutter comes back fine.

the panel on the right is the capture. the map behind it is rendering perfectly, and the capture of it is white.

### what we did about it

we have this working in a fork: the map is created without a platform view, hosted offscreen, and its frames are handed to `FlutterTextureRegistry`, so the widget tree gets a `Texture` and nothing else. annotations, the full `MapboxMap` api, pan, pinch, rotate, pitch, resize and the logo and attribution all work, and captures contain the map.

happy to open the pr. it is about 700 lines, three new files plus 20 lines touched in existing ones, and nothing changes for anyone using `MapWidget`.

before asking you to review that, two questions:

1. is an ios texture path something you would take in principle, or is uikit composition a deliberate constraint?
2. if you would take it, do you want it as a separate widget or as an option on `MapWidget`?

Nixxx19:
closes #1142

on ios the map is a `UiKitView`, so uikit composites it and it is not in flutter's scene. nothing can read it back: `toImageSync` and `RepaintBoundary.toImage` return a hole where the map is, and a custom `ImageFilter.shader` used as a backdrop samples an empty texture. android avoids this by rendering into a `TextureView`; ios has no equivalent because its platform views are always uikit composited (flutter/flutter#43902).

this adds a texture path for ios. the map is created without a platform view and its frames go to `FlutterTextureRegistry`, so the widget tree contains a `Texture` and nothing else.

```dart
MapTexture(
  styleUri: MapboxStyles.MAPBOX_STREETS,
  onMapCreated: (map) async {
    await map.setCamera(...);
    await map.style.addLayer(...);
  },
)
```

## what it gives you

* the map can be captured. screenshots, thumbnails and share cards that include the map now work.
* custom shaders and backdrop filters can sample the map, so glass, scrims and blurs driven by map luminance behave like they do anywhere else.
* the map composites like any other widget: transforms, opacity, clips, widgets between layers.
* **it is faster than the platform view** on the numbers below.
* everything already in the plugin keeps working through it: the full `MapboxMap` api, style, camera, sources, layers, view annotations, and the logo and attribution.

## before and after

same page, same button, same code, both on an iphone. the panel on the right is the result of `RepaintBoundary.toImage` on the page.

| `MapWidget` (platform view) | `MapTexture` |
|---|---|
|  |  |
| the capture is blank. the map is not in flutter's scene, so there is nothing to capture. | the capture contains the map: streets, labels, scale bar, logo and attribution. |

## performance

same device, same camera path, profile mode, frame timings from `SchedulerBinding.addTimingsCallback`. a small widget repaints continuously during the run so flutter produces frames to measure.

| | `MapWidget` (platform view) | `MapTexture` |
|---|---|---|
| frames | 136 | 128 |
| build p50 | 0.35 ms | **0.31 ms** |
| build p90 | 0.43 ms | **0.40 ms** |
| raster p50 | 0.74 ms | **0.56 ms** |
| raster p90 | 0.96 ms | **0.70 ms** |

raster p90 is 27% lower with the texture. ios platform view compositing splits the scene into overlay layers around the view, and a texture is one more layer in a single scene, so the per frame copy costs less than the compositing it replaces. the example app ships the page that produces these numbers, so the result is reproducible rather than quoted.

## how it works

`HeadlessMapTexture` builds the ordinary `MapboxMapController` with a channel suffix and parks its view offscreen inside the app's own key window, so every pigeon api behaves exactly as it does for the platform view. `MapTexturePublisher` copies each finished frame into an IOSurface backed `CVPixelBuffer` and hands it to the texture registry. on the dart side `MapTexture` sizes the map to its constraints, forwards gestures and disposes the native map when it unmounts.

three details worth calling out, because none of them is obvious:

**frames are caught at `-[MTLCommandBuffer presentDrawable:]`.** `UIView.draw(_:)` looks like the hook and never runs, because an MTKView renders through metal rather than core graphics. presentDrawable also lets the copy be encoded onto the same command buffer that presents it, so it is ordered after the render with no waiting.

**the offscreen map lives in the app's existing key window.** a second `UIWindow` takes over the scene. staying composited is what keeps coreanimation recycling drawables.

**the logo and attribution are rasterised and drawn by flutter.** they are uikit views drawn with core graphics, so they are not in the metal output. attribution is preserved.

## gestures

pan, pinch, rotate and pitch are forwarded and drive the camera directly, pan through `dragCameraOptions` which is what the sdk's own recogniser uses.

## scope

`MapWidget` is untouched. `MapTexture` is additive, ios only and opt in. the 20 lines changed in existing files are an export, a `part`, a `MapboxMap.headless` factory and a getter on `MapboxMapController` so it can be hosted outside a platform view. nothing existing changes behaviour.

## tests and verification

* 8 widget tests in `test/map_texture_test.dart`: creation at the size of its constraints, the texture rendering once the host returns an id, a unique channel suffix per map, style pass through, resize on a constraint change, disposal of the native map when it leaves the tree, gesture forwarding, and nothing forwarded when gestures are off
* `flutter analyze` clean on every touched file
* two example pages: the capture comparison and the performance comparison
* verified on an iphone: capture, annotations, pan, pinch, rotate, pitch, resize and disposal

## follow ups

happy to take these in this pr or later, whichever you prefer:

* the frame copy is a gpu blit. rendering straight into the flutter buffer is the obvious next step and would remove it.
* fling momentum on pan.

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -35,6 +35,7 @@ import 'package:mapbox_maps_flutter/mapbox_maps_flutter.dart';
 import 'full_map_example.dart';
 import 'location_example.dart';
 import 'example.dart';
+import 'map_texture_example.dart';
 import 'point_annotations_example.dart';
 import 'projection_example.dart';
 import 'rainbow_road_example.dart';
@@ -45,6 +46,7 @@ import 'map_recorder_example.dart';
 
 final List<Example> _allPages = <Example>[
   SimpleMapExample(),
+  MapTextureExample(),
   ViewportExample(),
   SnapshotterExample(),
   TrafficRouteLineExample(),
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -36,6 +36,7 @@ import 'full_map_example.dart';
 import 'location_example.dart';
 import 'example.dart';
 import 'map_texture_example.dart';
+import 'map_texture_perf_example.dart';
 import 'point_annotations_example.dart';
 import 'projection_example.dart';
 import 'rainbow_road_example.dart';
@@ -47,6 +48,7 @@ import 'map_recorder_example.dart';
 final List<Example> _allPages = <Example>[
   SimpleMapExample(),
   MapTextureExample(),
+  MapTexturePerfExample(),
   ViewportExample(),
   SnapshotterExample(),
   TrafficRouteLineExample(),
--- a/example/lib/map_texture_example.dart
+++ b/example/lib/map_texture_example.dart
@@ -1,0 +1,89 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/rendering.dart';
+import 'package:mapbox_maps_flutter/mapbox_maps_flutter.dart';
+import 'dart:ui' as ui;
+import 'example.dart';
+
+/// Shows the difference a texture makes: pressing the button captures this
+/// page with `RepaintBoundary.toImage`. With [MapTexture] the capture contains
+/// the map, because the map is ordinary flutter content. With a [MapWidget]
+/// the same capture comes back with a hole where the map is, because a
+/// platform view is composited by uikit and is not in flutter's scene.
+class MapTextureExample extends StatefulWidget implements Example {
+  @override
+  final Widget leading = const Icon(Icons.photo_camera_back);
+  @override
+  final String title = 'Map in a flutter texture (iOS)';
+  @override
+  final String? subtitle = 'No platform view, so the map can be captured';
+
+  @override
+  State<StatefulWidget> createState() => MapTextureExampleState();
+}
+
+class MapTextureExampleState extends State<MapTextureExample> {
+  final GlobalKey _boundary = GlobalKey();
+  ui.Image? _capture;
+
+  Future<void> _onMapCreated(MapboxMap map) async {
+    await map.setCamera(
+      CameraOptions(
+        center: Point(coordinates: Position(-0.0880, 51.5140)),
+        zoom: 12.5,
+      ),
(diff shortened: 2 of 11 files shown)


luizvnegrini:
awesome

